PR #4, opened 2020-05-10
Added in Course Flow

gbhojraj:
This PR adds in a chunk of the views for the Courses portion of a Professor's/Student's view.

Opening this PR without all functionality so it can be incrementally reviewed.


src/Courses/CourseDetailsTabs/CourseUsersList.tsx
```diff
@@ -0,0 +1,60 @@
...
+    setActive: React.Dispatch<React.SetStateAction<Set<number>[]>>;
+}
+
+export const CourseUsersList: React.FC<CourseUsersListProps> = ({users, activeUsers, setActive}) => {
+    // TODO: This should be refactored into a reusable TwoState button.
+    const [selectAllState, setSelectAllState] = useState(false);
+
+    // Add or remove a user's ID to the active users array.
+    const onClickStudent = (id: number) => {
+        if (activeUsers[0].has(id)) {
```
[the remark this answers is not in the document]
tommylettieri: If so why ^

src/Courses/CourseDetailsTabs/CourseUsersList.tsx
```diff
@@ -0,0 +1,60 @@
...
+        }
+    };
+
+    const onSelectAll = () => {
+        // If all are currently selected, select none.
+        if (selectAllState) {
+            setActive([new Set()]);
+        } else {
+            const allUserIds: Array<number> = users.map(user => user.id);
+            setActive([new Set(allUserIds)]);
```
tommylettieri, line 32: to avoid double array should we do `...allUserIds`

src/Courses/CourseDetailsTabs/CourseUsersList.tsx
```diff
@@ -0,0 +1,60 @@
...
+                        <FormControl type="text" />
+                    </Col>
+                    <Col md={2}>
+                        <Button onClick={() => onSelectAll()}>{selectAllState ? 'Unselect All' : 'Select All'}</Button>
+                    </Col>
+                </Row>
+            </Form.Group>
+            <ListGroup>
+                {users.map((user, i) => (
+                    <ListGroupItem key={i} active={activeUsers[0].has(user.id)} onClick={() => onClickStudent(user.id)}>
```
tommylettieri, line 52: index as key is bad practice and results in warning in console

tommylettieri: can't we just use user.id

src/Courses/CoursePage.tsx
```diff
@@ -1,22 +1,51 @@
...
-        setCourses(['Course 1', 'Course 2', 'Course 3']);
+        (async () => {
+            try {
+                let res = await AxiosRequest.get('/courses');
+                console.log(res.data.data);
+                const courses: Array<CourseObject> = map(res.data?.data, obj => new CourseObject(obj));
+
+                setCourses(courses);
+            } catch (e) {
+                console.log(e.response);
```
tommylettieri, line 28: should we have `console.log`
should we use a client side logger?

src/Courses/CoursePage.tsx
```diff
@@ -1,22 +1,51 @@
...
         <div>
-            <h1>My Courses</h1>
-            <CourseList courses={courses}/>
+            <Container>
+                <Row>
+                    <Col md={10}>
+                        <h1>My Courses</h1>
+                    </Col>
+                    <Col md={2}>
+                        {userType === 'Professor' && (
```
tommylettieri, line 42: We should have user type in an enum and compare that

src/Courses/CourseRowComponent.tsx
```diff
@@ -0,0 +1,21 @@
...
+export const CourseRowComponent: React.FC<CourseObject> = ({ course_name, course_start, course_end}) => {
+    const COURSE_DATE_FORMAT = 'MMM D, YYYY';
+
+    return (
+        <>
+            <Row>
+                {course_name}
+            </Row>
+            <Row>
+                {/* TODO: Timezone support? */}
```
tommylettieri, line 15: should localize, shouldn't matter, we should just verify that is the case

src/Courses/TopicsList.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+/**
+ * 
+ * @param param0 
+ */
+export const TopicsList: React.FC<TopicsListProps> = ({listOfTopics}) => {
+    return (
+        <ListGroup>
+            {listOfTopics.map((topic, i) => (
+                <ListGroupItem key={`topic-${i}`}>
```
tommylettieri, line 16: no indexes, topic.id?

src/Courses/CourseDetailsPage.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Container, Tabs, Tab } from 'react-bootstrap';
+import EnrollmentsTab from './CourseDetailsTabs/EnrollmentsTab';
+import TopicsTab from './CourseDetailsTabs/TopicsTab';
+
+interface CourseDetailsPageProps {
+
+}
+
+enum CourseDetailsTabs {
+    // eslint-disable-next-line no-unused-vars
```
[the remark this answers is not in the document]
gbhojraj: No, eslint doesn't seem to detect that these enum values are actually being used.

src/Courses/CourseDetailsPage.tsx
```diff
@@ -0,0 +1,37 @@
...
+    TOPICS = 'topics',
+    // eslint-disable-next-line no-unused-vars
+    ENROLLMENTS = 'enrollments'
+}
+
+/**
+ * This page renders a tabbed view of course details. If a user is a professor, this will have an additional tab
+ * to view enrolled students and send emails.
+ * @param param0
```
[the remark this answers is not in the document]
gbhojraj: Ah, that's some extra data from the autocomplete when I make a block comment. I leave it in if I want to go back and add a parameter description.

src/Courses/CourseDetailsTabs/CourseUsersList.tsx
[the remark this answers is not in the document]
gbhojraj: This is a bit of a hack to use a Set, which provides implicit uniqueness, within state. Because the Set is an object, state won't update when it updates because the reference won't change. Creating an array each time with a set inside it will cause the state to update.

src/Courses/CourseDetailsTabs/CourseUsersList.tsx
gbhojraj: The `Set` constructor accepts an array.

src/Courses/CourseDetailsTabs/CourseUsersList.tsx
gbhojraj: It doesn't throw an error, but you're right, I forgot about that caveat with keys. I'll fix this.

src/Courses/CoursePage.tsx
gbhojraj: I think for now, it's okay. I don't think I've used any client-side loggers, but I'll look into `loglevel` and anything you might think up.

src/Courses/CoursePage.tsx
gbhojraj: I'll handle this in a new PR- I have to actually hook in the role, and right now I'm just hard-coding it.

src/Courses/CourseDetailsPage.tsx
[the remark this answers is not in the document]
gbhojraj: I fixed this by turning off the eslint rule and adding in the typescript equivalent explicitly.

src/Courses/CourseDetailsTabs/CourseUsersList.tsx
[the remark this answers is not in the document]
tommylettieri: I would rather we had an object that was described in an interface with a data property of type any and then update that, I think activeUsers.data is easier to read then activeUsers[0]

PR #15, opened 2020-05-25
Feature/add def upload

gbhojraj:


src/Courses/CourseCreation/CourseEditPage.tsx
```diff
@@ -1,41 +1,73 @@
...
+            let course = await AxiosRequest.get(`/curriculum/${courseId}`);
+
+            console.log(course.data.data);
+            setCourse(course.data.data);
+        })();
+    }, []);
+
+
+    const callShowTopicCreation = (unit: number, e: any = null) => {
+        if (e != null) {
```
tommylettieri, line 36: why single equals?

tommylettieri: null and undefined I would guess, but does this also catch things like false?  Even though it should be fine I feel like we should keep up with good practice, doesn't linter yell about this?

src/Courses/CourseDetailsPage.tsx
```diff
@@ -18,14 +21,62 @@ enum CourseDetailsTabs {
...
         <Container>
-            <Tabs defaultActiveKey={CourseDetailsTabs.TOPICS} id="course-details-tabs">
+            <Tabs defaultActiveKey={CourseDetailsTabs.DETAILS} id="course-details-tabs">
+                <Tab eventKey={CourseDetailsTabs.DETAILS} title="Details">
+                    {course && (
+                        <>
+                            <h1>{course.name}</h1>
+                            <p>Course description content goes here.</p>
+                            <Button onClick={() => copyToClipboard()}>Copy Student Enrollment Link</Button>
+                            <a ref={a => enrollLinkRef = a} id='enrollUrlLink' href={enrollUrl}>https://{enrollUrl}</a>
```
tommylettieri, line 70: hardcoding https?

src/Courses/CourseDetailsTabs/TopicsTab.tsx
```diff
@@ -1,32 +1,29 @@
...
-    {name: 'Unit 3', topics: mock_topics, unit_id: 3},
-];
-
 export const TopicsTab: React.FC<TopicsTabProps> = () => {
+    const { courseId } = useParams();
+    const [course, setCourse] = useState<any>(null);
+
+    useEffect(() => {
+        (async ()=>{
+            let course = await AxiosRequest.get(`/curriculum/${courseId}`);
```
tommylettieri, line 17: Is there any reason we didn't separate server integration into it's own files

gbhojraj: What do you mean, `server integration`? Are you asking why we don't pull out all the API requests into a separate file? We could have created named functions to pull in and pass into useEffect, but I'm not sure there's an easy way to handle state with that.

src/Courses/CourseCreation/CourseEditPage.tsx
gbhojraj: Using soft equals is fine where it's intentional. In this case, I want any falsy value to fail. e should be an event, anyhow.

src/Courses/CourseDetailsPage.tsx
gbhojraj: Is this a problem?

PR #18, opened 2020-05-30
Feature/add topic creation

gbhojraj:


src/Courses/CourseCreation/CourseEditPage.tsx
```diff
@@ -1,51 +1,239 @@
...
+            }
+
+            postObject.curriculumId = parseInt(courseId, 10);
+            console.log('Creating a new course', postObject);
+            return await AxiosRequest.post('/courses', postObject);
+        };
+
+        let res = await createCourse(course);
+        console.log(res);
+        if (res?.status !== 201) {
```
tommylettieri, line 136: I would just let axios do the error handling, by default it checks between [200, 300)

src/Courses/CourseDetailsTabs/EnrollmentsTab.tsx
```diff
@@ -1,21 +1,25 @@
...
-    new UserObject({first_name: 'Henry', last_name: 'McCoy', id: 2}),
-    new UserObject({first_name: 'Jean', last_name: 'Grey', id: 3}),
-    new UserObject({first_name: 'Anne', last_name: 'LeBeau', id: 4}),
-];
+export const EnrollmentsTab: React.FC<EnrollmentsTabProps> = ({courseId}) => {
+    const [users, setUsers] = useState([]);
+
+    useEffect(() => {
+        (async () => {
+            const usersResp = await AxiosRequest.get(`users?courseId=${courseId}`);
```
tommylettieri, line 15: We could just pass courseId in the params of the options, doesn't really matter in this case because url encoding will never be an issue and there is only one parameter but for consistent in case that isn't the case in other request I'd prefer params

PR #22, opened 2020-06-06
Feature/move def upload to topic creation

gbhojraj:


drdrew42:
problems encountered with "save course" - though my guess is that they are unrelated to this feature. Adding "too many" problems at once seems to be the source of the issue.
Screen Shot 2020-06-06 at 10 36 18 AM

PR #24, opened 2020-06-07
Feature/refactor problem page

gbhojraj:


src/Courses/CourseCreation/TopicCreationModal.tsx
```diff
@@ -133,10 +134,27 @@ export const TopicCreationModal: React.FC<TopicCreationModalProps> = ({unit,  ad
...
+        const problemsWithOrdering = problems.map((problem, index) => {
+            // Problems should always render in the order that the professor sets them.
+            problem.problemNumber = index+1;
+
+            // If we don't recognize the prefix, assume they're using Contrib.
+            if (!_.startsWith(problem.webworkQuestionPath, 'Contrib') || !_.startsWith(problem.webworkQuestionPath, 'OpenProblemLibrary')) {
+                problem.webworkQuestionPath = `webwork-open-problem-library/Contrib/${problem.webworkQuestionPath}`;
+            } else {
+                problem.webworkQuestionPath = `webwork-open-problem-library/${problem.webworkQuestionPath}`;
+            }
```
drdrew42, line 148: Something about this logic isn't working properly - I'm importing def files that start with Contrib, but they're ending up as `webwork-open-problem-library/Contrib/Contrib/...`

*also* def files (almost) ALWAYS use `Library` as a symlink for `OpenProblemLibrary` -- this is the default for WeBWorK, and we should assume that everyone has observed this standard.

so we want to replace Library with OpenProblemLibrary;
we want to leave Contrib as is;
and _everything_ gets `webwork-open-problem-library` prepended
(I have fixed all my def files that skipped Contrib/ and started at CUNY/...)

drdrew42: fixed the logic, and added the "Library" conversion.

Issue #26, opened 2020-06-12
Bug: iFrame doesn't resize after submit


gbhojraj:
The Problem iFrame doesn't resize after a successful submit. This is because of a CORS issue. We can resolve this by either (1) hijacking the POST and using the srcDoc, or (2) looking into fixing the CORS-causing references in the rendered HTML.

gbhojraj:
https://stackoverflow.com/questions/42746737/how-to-hijack-html-form-input-and-accurately-convert-the-data-for-use-in-ajax-re

PR #27, opened 2020-06-12
Feature/hijack form submit

gbhojraj:


gbhojraj:
This should take care of #26 , since we're using srcDoc to get around the CORS problem.

src/Courses/CourseDetailsPage.tsx
```diff
@@ -73,7 +75,7 @@ export const CourseDetailsPage: React.FC<CourseDetailsPageProps> = () => {
                                     aria-label="Enrollment link"
                                     aria-describedby="basic-addon2"
                                     ref={textAreaRef}
-                                    value={`https://${enrollUrl}`}
+                                    value={`http://${enrollUrl}`}
```
tommylettieri, line 78: Yeah I noticed this last night, it would be nice if we could get the protocol from the browser

PR #40, opened 2020-06-21
This is a temporary handling of the submit button functionality in th…

ghost:
…e iFrame, until we can move them out of the iFrame.


drdrew42:
What is the effect if we attempt to _.omit keys that aren't present in the form-data?
For example, the GeoGebra Applet problem that we've been testing uses an optional "return this problem to its original state" button. (This option is available in all problems that use applets, some use it, others don't.) Because of this, "previewAnswers" is getting through the form scrubber because it's not listed.

drdrew42:
Polyfill for browsers that do not support `e.submitter`
(I'm not sure how to go about including this through typescript...)

```javascript
!function(){
    var lastBtn = null
    document.addEventListener('click',function(e){
        if (!e.target.closest) return;
        lastBtn = e.target.closest('button, input[type=submit]');
    }, true);
    document.addEventListener('submit',function(e){
        if ('submitter' in e) return;
        var canditates = [document.activeElement, lastBtn];
        for (var i=0; i < canditates.length; i++) {
            var candidate = canditates[i];
            if (!candidate) continue;
            if (!candidate.form) continue;
            if (!candidate.matches('button, input[type=button], input[type=image]')) continue;
            e.submitter = candidate;
            return;
        }
        e.submitter = e.target.querySelector('button, input[type=button], input[type=image]')
    }, true);
}();
```

gbhojraj:
I'm thinking that we don't need to merge this since we're going to have the new renderer template.

tommylettieri:
We do need to implement this check, even with the new template. For example, GeoGebra applet problems have an option to include a “reset to original state” button with the applet. This button is named “previewAnswers” (so that when pressed, WeBWorK does not grade it as an attempt). I do not have a good means to block this button on the renderer side (without deviating from the “standard” PG library).
TL;DR problems can still introduce form buttons into their code that are not part of the template.

PR #43, opened 2020-07-05
Feature/add backend support for grades

ghost:


src/Courses/CourseDetailsTabs/GradeTable.tsx
```diff
@@ -1,31 +1,64 @@
...
 import _ from 'lodash';
+import MaterialTable from 'material-table';
+import { Clear, SaveAlt, FilterList, FirstPage, LastPage, ChevronRight, ChevronLeft, Search, ArrowDownward } from "@material-ui/icons";
 
 interface GradeTableProps {
+    courseName: string;
     grades: Array<any>;
     onRowClick?: (id: number) => void;
 }
+const icons = {
```
tommylettieri, line 13: Should we move this somewhere global so we can reuse the icons?

PanopticaRising: I'll resolve this in a new PR when I add the other Material UI component.

PR #60, opened 2020-07-24
Feature/add polyfills

PanopticaRising:


src/Courses/CourseCreation/TopicCreationModal.tsx
```diff
@@ -165,7 +213,7 @@ export const TopicCreationModal: React.FC<TopicCreationModalProps> = ({unitIndex
             <Modal.Header closeButton>
                 <h3>{existingTopic ? `Editing: ${existingTopic.name}` : 'Add a Topic'}</h3>
             </Modal.Header>
-            <Modal.Body>
+            <Modal.Body style={{minHeight: `${24 + (problems.length * 19)}vh`}}>
```
red-tt-bear, line 216: Calculating the style and using "view height" should probably be avoided

red-tt-bear: I can see the argument that it's a modal so it's ok

PR #71, opened 2020-07-25
Bugfix/fix datepicker on course creation

PanopticaRising:


doombot-exe:
Something is now off with "correct" "in progress" "incorrect" - the tags are being applied to the problem _after_ the current problem.

doombot-exe:
units now added, topics now added, still issues:

* empty topic is allowed - "error loading this assignment" (because empty)
* new topics are created starting with problem 0 
* minimal validation to make sure problem path regex 'm/[a-zA-Z0-9 _-\.\/]*.pg/'

doombot-exe:
datepicker during course creation works for course start/end and for topic start/end

PR #75, opened 2020-07-27
Feature/show open topics on details page

PanopticaRising:


PanopticaRising:
@red-tt-bear This currently doesn't work if you click the topic to launch into the problem screen because the GetTopics?isOpen call doesn't pass back the questions. I can try to map this against what we already have from the course if you don't want to join the questions in that call.

red-tt-bear:
> @red-tt-bear This currently doesn't work if you click the topic to launch into the problem screen because the GetTopics?isOpen call doesn't pass back the questions. I can try to map this against what we already have from the course if you don't want to join the questions in that call.

To hyperlink them to our current questions screen? That seems like a totally different url, can't we just navigate to the other component?

doombot-exe:
The point would be to land the students _in_ the problem set if they click from the details page -- the same as though they'd clicked on the topic from the usual units/topics page

red-tt-bear:
I see the problem your talking about, there is a different issue here:
The component does not fetch the data it is passed in, however we have a url (i.e. http://localhost:33002/common/courses/2/1) which does not work when you enter it directly)
The real issue here is we need a backend call to get problems by topic right?

Issue #45, opened 2020-07-05
Add Accessibility


gbhojraj:
Accessibility is hugely important for the platform. We currently have the ability to audit with a11y plugins and the WAVE Chrome extension.

gbhojraj:
This branch adds in a11y checking:
https://github.com/rederly/frontend/tree/feature/a11y-audit

PR #101, opened 2020-08-01
Feature/a11y audit

PanopticaRising:


PanopticaRising:
This should be good eneough to make sure we address #45

src/Router.tsx
```diff
@@ -3,6 +3,7 @@ import { BrowserRouter, Switch, Route } from 'react-router-dom';
 import HomePage from './Home/HomePage';
 import VerificationLandingPage from './Login/VerificationLandingPage';
 import NavWrapper from './NavWrapper/NavWrapper';
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
```
red-tt-bear, line 6: We should comment out instead of ignoring the error

src/index.tsx
```diff
@@ -4,6 +4,11 @@ import './index.css';
 import * as serviceWorker from './serviceWorker';
 import { Router } from './Router';
 
+if (process.env.NODE_ENV !== 'production') {
```
red-tt-bear, line 7: Is this guarenteed?

src/Router.tsx
red-tt-bear: @PanopticaRising said that if a file isn't included it throws errors which is why it is here
We should leave a comment for this
Better yet we could just do `import './Adviser/AdviserPage';`

src/index.tsx
doombot-exe: yes. 
https://create-react-app.dev/docs/adding-custom-environment-variables/

PR #109, opened 2020-08-13
Feature/red 41 students can see own grades like statistics

PanopticaRising:


src/Courses/CourseList.tsx
```diff
@@ -13,11 +13,16 @@ export const CourseList: React.FC<CourseListProps> = ({courses}) => {
 
     return (
         <ListGroup>
-            {courses.map(obj => <ListGroupItem action href={`${url}/${obj.id}`} key={`course${obj.id}`}>
-                <CourseRowComponent 
-                    {...obj}
-                />
-            </ListGroupItem>)}
+            {/* TODO: Make a link tag */}
```
red-tt-bear, line 16: It is a link?

src/Courses/EnrollUserPage.tsx
```diff
@@ -37,11 +41,13 @@ export const EnrollUserPage: React.FC<EnrollUserPageProps> = () => {
...
     return (
         <Jumbotron>
-            {enrollError && <h2>{enrollError}</h2>}
-            {enrollData && <>
-                <h4>You have been enrolled in this class!</h4>
-                <h3><Link to={`/common/courses/${enrollData.courseId}`}>Visit the class now.</Link></h3>
-            </>}
+            {enrollError ? 
+                <h2>{enrollError}</h2> :
+                (enrollData.courseId > -1 ? <>
```
red-tt-bear, line 46: Due to this case I think we should make enrollData optional

src/Courses/CourseList.tsx
PanopticaRising: Yeah, this was my way of remembering to do it.

PR #111, opened 2020-08-17
Feature/red 15 hook prof grades to student grades

PanopticaRising:


src/Courses/CourseDetailsPage.tsx
```diff
@@ -34,7 +35,9 @@ export const CourseDetailsPage: React.FC<CourseDetailsPageProps> = () => {
     const { courseId } = useParams();
     const [course, setCourse] = useState<CourseObject>(new CourseObject());
     const [activeTab, setActiveTab] = useState<CourseDetailsTabs>(CourseDetailsTabs.DETAILS);
-    const userType: UserRole = getUserRole(Cookies.get(CookieEnum.USERTYPE));
+    const userType: UserRole = getUserRole();
+    const userId: number = getUserId();
+    const [studentNameAndId, setStudentNameAndId] = useState<{name: string, userId: number} | null>(null);
```
red-tt-bear, line 40: I don't like that we have two attributes in the same hook, if one were to change it wouldn't affect the other, granted taht shouldn't happen but I don't know how much of an advantage this really gives you

red-tt-bear: Follow up: would it better to use a 'me' parameter on the backend?

src/Enums/UserRole.tsx
```diff
@@ -39,4 +39,19 @@ export const getUserRole = (roleFromCookie: string | undefined): UserRole => {
     default:
         return UserRole.STUDENT;
     }
+};
+
+export const getUserId = () => {
+    const userId = Cookies.get(CookieEnum.USERID);
+    let userIdValue = 0; 
+
+    if (userId === undefined || isNaN(userIdValue = parseInt(userId, 10))) {
```
red-tt-bear, line 48: should we be using `_.isNil` instead of `=== undefined`
Is that just for isStrict? because `isNaN` should catch that

PR #114, opened 2020-08-23
Added breadcrumbs to the NavWrapper.

PanopticaRising:


src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -161,7 +161,7 @@ export const StatisticsTab: React.FC<StatisticsTabProps> = ({course, userId}) =>
                 return;
             }
         })();
-    }, [course.id, view, idFilter, userId]);
+    }, [course, course.id, userType, view, idFilter, userId]);
```
red-tt-bear, line 164: Why are we checking course, if the id doesn't change we shouldn't make the request

red-tt-bear: I think this makes extra calls to the backend

PR #150, opened 2020-09-14
Feature/red 9 grade override

red-tt-bear:


src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -24,18 +31,34 @@ enum StatisticsView {
     ATTEMPTS = 'ATTEMPTS',
 }
 
+enum StatisticsViewFilter {
+    UNITS_FILTERED = 'UNITS_FILTERED',
```
PanopticaRising, line 35: You should use `UNITS_${FILTERED_STRING}` so that it remains consistent, but it's not a big deal.

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -69,51 +92,90 @@ const icons = {
 };
 
 
+interface BreadCrumbFilter {
+    id: number;
+    displayName: string;
+}
+
+type EnumDictionary<T extends string | symbol | number, U> = {
```
PanopticaRising, line 100: Should this be moved to a Utils file?

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -69,51 +92,90 @@ const icons = {
...
             idFilterLocal = null;
             break;
         default:
-            url = `${url}/units?`;
-            filterParam = '';
-            if(idFilterLocal !== null) {
-                console.error('This should be null for units');
-                idFilterLocal = null;
-            }
+            console.error('You should not havea  view that is not the views or filtered views');
```
PanopticaRising, line 173:
```suggestion
            console.error('You should not have a view that is not the views or filtered views');
```

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -162,83 +227,220 @@ export const StatisticsTab: React.FC<StatisticsTabProps> = ({course, userId}) =>
...
-                    </Nav.Link>
-                </Nav.Item>
-                {userId !== undefined && <Nav.Item>
-                    <Nav.Link eventKey={StatisticsView.ATTEMPTS}>
-                        Attempts
-                    </Nav.Link>
-                </Nav.Item>}
+                {Object.keys(StatisticsView).map((key: string) => (
+                    (key !== StatisticsView.ATTEMPTS || userId !== undefined) &&
+                    <Col key={`global-${key}`} className="p-0" >
```
PanopticaRising, line 378: What is this classname?

PanopticaRising (approved):

src/Courses/CourseDetailsTabs/OverrideGradeModal.tsx
```diff
@@ -0,0 +1,250 @@
...
+                                            size='lg'
+                                            readOnly={loading}
+                                            type='number'
+                                            step={0.1}
+                                            min={0}
+                                            // TODO should we put a max or is it a feature if there is no max
+                                            // max={100}
+                                            onChange={onNewScoreChange}
+                                        />
+                                        <Form.Control.Feedback type="invalid">{<span>The new score must be a postive number between 0 and 100</span>}</Form.Control.Feedback>
```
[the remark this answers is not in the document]
red-tt-bear: That's a good point, let's talk about in scrum if this is a bug or a feature

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
red-tt-bear: Yeah I added the string for _FILTERED after the fact
The problem is I want the value to match the key, not sure if  enums will let me use variables in the key

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
red-tt-bear: Yes, I meant to do that 😆

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
red-tt-bear: padding 0, instead of doing inline styles we should probably use bootstrap classes more, even if we move away from react bootstrap

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
red-tt-bear: I get this error if I try to put it in the key: `Computed property names are not allowed in enums.ts`
So the question is do we want the key and value to be a strict copy and paste or do we want the value to use the constant 🤔

PR #151, opened 2020-09-15
Bugfix/fix spacing on topic tab buttons

PanopticaRising:


package.json
```diff
@@ -47,7 +47,8 @@
   "scripts": {
     "start": "PORT=3002 react-scripts start",
     "server": "npm run start:watch --prefix ../backend/",
-    "renderer": "(cd ../renderer-docker && ./start_docker.sh)",
+    "renderer": "(cd ../renderer && sudo morbo ./script/render_app)",
```
red-tt-bear, line 50: I don't like this, there are assumptions about your system, I would rather this was in a parent project with submodules

src/Components/EditToggleButton.tsx
```diff
@@ -11,15 +12,27 @@ interface EditToggleButtonProps {
     onClick?: (newState: boolean, oldState: boolean)=>void;
 }
 
+const selectedButton = (
```
red-tt-bear, line 15: I go back and forth on inlining vs not inlining... It's nice to be inlined for readability, but it's better to be up top for reuse

red-tt-bear (approved):

PR #152, opened 2020-09-16
Added support for Logrocket to get data on user sessions.

PanopticaRising:


src/Login/LoginForm.tsx
```diff
@@ -56,6 +57,12 @@ export const LoginForm: React.FC<LoginFormProps> = () => {
                 Cookie.set(CookieEnum.USERID, resp.data.data.userId);
                 Cookie.set(CookieEnum.USERNAME, `${resp.data.data.firstName} ${resp.data.data.lastName}`);
 
+                LogRocket.identify('qdgaa4/rederly', {
+                    name: `${resp.data.data.firstName} ${resp.data.data.lastName}`,
```
red-tt-bear, line 61: @doombot-exe do we have to worry about FERPA or private info here at all, I don't think this is an issue

src/index.tsx
```diff
@@ -3,11 +3,17 @@ import ReactDOM from 'react-dom';
...
 import { Router } from './Router';
+import LogRocket from 'logrocket';
 
-if (process.env.NODE_ENV !== 'production') {
-    var axe = require('react-axe');
-    axe(React, ReactDOM, 1000);
-}
+LogRocket.init('qdgaa4/rederly');
+
+// This causes slowdowns and adversely affects development.
```
red-tt-bear, line 10: Thank you! :D

src/Login/LoginForm.tsx
doombot-exe: Talking about this now actually

PR #153, opened 2020-09-16
Added the URL to the tail-end of the description for the provide feedback

PanopticaRising:
@red-tt-bear I realize this is probably the short and messy way to do it. I could add a new query parameter to the backend instead.

### DIFF
--- a/src/NavWrapper/ProvideFeedback.tsx
+++ b/src/NavWrapper/ProvideFeedback.tsx
@@ -1,5 +1,6 @@
 import React, { useState } from 'react';
 import { Button, Modal, FormControl, FormLabel, FormGroup, Spinner, Form, } from 'react-bootstrap';
+import { useHistory } from 'react-router-dom';
 import AxiosRequest from '../Hooks/AxiosRequest';
 
 export const ProvideFeedback: React.FC<any> = () => {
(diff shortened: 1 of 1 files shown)


src/NavWrapper/ProvideFeedback.tsx
```diff
@@ -10,14 +11,16 @@ export const ProvideFeedback: React.FC<any> = () => {
...
 
     const submitFeedback = async () => {
         try {
             setMessage('');
             setSubmitting(true);
             setEnabled(false);
+            console.log(history);
             await AxiosRequest.post('/support', {
-                description,
+                description: `${description} (Sent from: ${window.location.origin}${history.location.pathname})`,
```
red-tt-bear, line 23: This should be a separate field and should be implemented on the backend
We might want to put this in a separate field in the future and I want all jira configurations on the backend

src/NavWrapper/ProvideFeedback.tsx
```diff
@@ -10,14 +11,16 @@ export const ProvideFeedback: React.FC<any> = () => {
...
     const [enabled, setEnabled] = useState(true);
     const [validated, setValidated] = useState(false);
+    const history = useHistory();
 
     const submitFeedback = async () => {
         try {
             setMessage('');
             setSubmitting(true);
             setEnabled(false);
+            console.log(history);
```
red-tt-bear, line 21: We should get rid of console log history

src/NavWrapper/ProvideFeedback.tsx
red-tt-bear: Made a ticket for fixing this, we don't need to block merging for this, extra info is good and I want it for the release:
https://rederly.atlassian.net/browse/RED-124

red-tt-bear (approved):

PR #154, opened 2020-09-16
Feature/red 9 grade override

red-tt-bear:


src/Assignments/ProblemDetails.tsx
```diff
@@ -168,28 +168,49 @@ export const ProblemDetails: React.FC<ProblemDetailsProps> = ({
...
-                        if (grade.overallBestScore >= 1) {
+                        if (_.isNil(problem)) {
+                            // The user should never see this
+                            // TODO maybe problem shouldn't be able to be null?
+                            message = applicationError;
+                        }
+                        // ******************** NOT RECORDING ********************
+                        else if (_.isNil(grade)) {
+                            if (getUserRole() === UserRole.STUDENT) {
+                                message = 'There is no grade associated with this problem. Your attempts will not be recorded.';
```
PanopticaRising, line 183: This seems a little strange to be a user-facing message, doesn't it? If you didn't know that a grade object had to exist in the database? I'm not sure what a better verbiage would be.

src/Assignments/ProblemDetails.tsx
```diff
@@ -168,28 +168,49 @@ export const ProblemDetails: React.FC<ProblemDetailsProps> = ({
...
-                        } else if (currentMoment.isBefore(deadDate) && currentMoment.isAfter(endDate)) {
-                            message = 'The topic is past due but partial credit is available. Your attempts will be graded with a penalty.';
                         } else if (currentMoment.isAfter(solutionsMoment)) {
-                            // TODO get from backend
+                            // TODO get solutionsMoment from backend
                             message = 'Solutions are available, your attempts will not be recorded.';
+                        }
+                        // ******************** RECORDING BUT NOT UPDATING GRADE \/\/\/\/********************
+                        else if (grade.locked === true) {
+                            message = 'Your grade on this problem has been locked. Your attempts will be recorded but your grade will not update. Clarify with your professor if you think this is an error.';
```
PanopticaRising, line 196: `Clarify` is a strange word to use here. I would suggest 
```suggestion
                            message = 'Your grade on this problem has been locked. Your attempts will be recorded but your grade will not update. Contact your professor if you think this is an error.';
```

doombot-exe (approved):

PanopticaRising (approved):

PR #157, opened 2020-09-18
Feature/allow professors to drop students

PanopticaRising:


src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
```diff
@@ -255,4 +256,21 @@ export const getQuestions = async ({
     } catch (e) {
         throw new BackendAPIError(e);
     }
+};
+
+export const deleteEnrollment = async ({
```
doombot-exe, line 261: do we bother with deleting the associated grades entries and further down to workbooks even? Logically, we should check for any workbooks/non-zero grades and warn about deleting data. Realistically, deleting an enrollment should apply to situations where (as has already happened) students accidentally get their hands on the wrong enrollment URL -- in which case, there should be no meaningful data generated in grades/workbooks.

doombot-exe: okay, this is not an actual delete - I misunderstood. This is just for dropping.

doombot-exe: this raises the question of whether or not we allow for _actual_ enrollment deletion, specifically reserved for cases where the enrollment is determined as invalid -- where the student should *not* preserve access to the course. When a student is "truly" enrolled and later drops/is dropped, they should retain access to the work they did in the course; however, for a student who "accidentally" enrolls, they should be removed and retaining access would be inappropriate...

doombot-exe (approved):

red-tt-bear:
Reviewed after the fact, looks good!

PR #158, opened 2020-09-19
Feature/red 2 save answer state

doombot-exe:
do not merge until after feature/RED-109-exams-fields is merged on the backend


src/Contexts/CurrentProblemState.tsx
```diff
@@ -0,0 +1,32 @@
...
+            lastSubmittedAt, 
+            setLastSavedAt, 
+            setLastSubmittedAt 
+        }}>
+            {children}
+        </ProblemStateContext.Provider>
+    );
+};
+
+export const useCurrentProblemState = () => React.useContext(ProblemStateContext);
```
red-tt-bear, line 32: @doombot-exe I don't think I like the idea of using context here, it seems like a lot of overhead (from a complexity standpoint and probably an efficiency standpoint however that is probably negligible) for passing the state up one level and down to a different level

@PanopticaRising do you have any thoughts on this?

src/Assignments/ProblemDetails.tsx
```diff
@@ -5,6 +5,7 @@ import moment from 'moment';
 import { OverlayTrigger, Tooltip, Badge } from 'react-bootstrap';
```
red-tt-bear, line 5: (This comment isn't actually for this line but rather the file)
This diff is massive, I think it's just things getting moved around but I would like to have a quick call to understand the changes here

It looks good to me thought

src/Assignments/ProblemIframe.tsx
```diff
@@ -71,6 +79,79 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
         setHeight(`${scrollHeight}px`);
     };
 
+    const formDataToObject = (formData: FormData) => {
+        let object:any = {};
+        // @ts-ignore
```
red-tt-bear, line 84: This was a temporary solution, we should see if there is a solution in which we don't have to ts-ignore

Also if we ts-ignore something we should put a comment above as to why we are doing so

src/Assignments/ProblemIframe.tsx
```diff
@@ -71,6 +79,79 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+        }
+        return object;
+    };
+
+    async function prepareAndSubmit(problemForm: HTMLFormElement, clickedButton?: HTMLButtonElement) {
+        const submitAction = (window as any).submitAction;
+        if(typeof submitAction === 'function') submitAction(); // this is a global function from renderer - prepares form field for submit
+
+        let formData = new FormData(problemForm);
+        if (_.isNil(problem.grades)) {return;} // just for typescript
```
red-tt-bear, line 103: This is just for typescript and should never occur, but since that is the case we should add some sort of log here
That way in the future when we clean up our logging we can have this send up to the server so we know when impossible cases are getting hit

src/Assignments/ProblemIframe.tsx
```diff
@@ -71,6 +79,79 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+        if (_.isNil(problem.grades[0])) {return;} // not enrolled
```
red-tt-bear, line 104: So this isn't an error, since you can submit questions when you are not enrolled

The valid use case for this is a professor trying their own problems

Does this prevent a professor from submitting their answers?

src/Assignments/ProblemIframe.tsx
```diff
@@ -71,6 +79,79 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+        if (_.isNil(problem.grades[0].id)) {
+            // PANIC -- should not happen
+            setError(`No grades id for problem #${problem.id}`);
```
red-tt-bear, line 107: Maybe we should do a console.error here as well or instead

Once we get the remote logging in I plan to do a global search for console.error to find where we need it

I'm thinking the remote logger will return a reference number and we can set the error to:
`Unexpected application error, contact rederly support with Rederly Reference Number #EXAMPLE_NUMBER` (since there is no corrective action for the user here)

src/Assignments/ProblemIframe.tsx
```diff
@@ -71,6 +79,79 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+            return;
+        }
+        const submiturl = _.isNil(clickedButton) ? `/backend-api/courses/question/grade/${problem.grades[0].id}` : problemForm.getAttribute('action');
```
red-tt-bear, line 110: We should remove this since it is an artifact of the previous implementation

We should do as we discussed and add a getter for the submit url and on form submit verify that the urls are the same

src/Assignments/ProblemIframe.tsx
```diff
@@ -82,50 +163,33 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
             // console.error('Hijacker: Could not find the form to insert the listener');
             return;
         }
-        problemForm.addEventListener('submit', (event: { preventDefault: () => void; }) => {
+
+        problemForm.addEventListener('submit', _.debounce((event: { preventDefault: () => void; }) => {
```
red-tt-bear, line 167: I thought we discussed moving the debounce to prepare and submit in this case since other buttons are considered submits (i.e. geogebra reset)

src/Assignments/ProblemIframe.tsx
```diff
@@ -82,50 +163,33 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
             let clickedButton = problemForm.querySelector('.btn-clicked') as HTMLButtonElement;
             if (_.isNil(clickedButton)) {
                 setError('Hijacker: An error occurred');
                 console.error('Could not find the button that submitted the form');
                 return;
             }
-            formData.set(clickedButton.name, clickedButton.value);
-            const submiturl = problemForm.getAttribute('action');
-            if(_.isNil(submiturl)) {
+            console.log('preparing formdata and submitting!');
```
red-tt-bear, line 180: I personally prefer not to have console logs in there

src/Courses/CourseInterfaces.tsx
```diff
@@ -129,6 +129,7 @@ export class StudentGrade {
     bestScore: number = 0;
     numAttempts: number = 0;
     locked: boolean = false;
+    currentProblemState?: JSON;
```
red-tt-bear, line 132: Interesting, I was unaware of a json data type, seems like something the linter wouldn't be happy with since it seems like another way to say `any`

src/Assignments/ProblemIframe.tsx
```diff
@@ -23,14 +26,16 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
-    readonly = false
+    readonly = false,
 }) => {
     const iframeRef = useRef<HTMLIFrameElement>(null);
     const [renderedHTML, setRenderedHTML] = useState<string>('');
     const [loading, setLoading] = useState(true);
     const [error, setError] = useState('');
     const [height, setHeight] = useState('100vh');
 
+    const { setLastSavedAt, setLastSubmittedAt } = useCurrentProblemState();
```
red-tt-bear, line 37: I'm assuming this does break the grades view of an attempt?

src/Assignments/ProblemIframe.tsx
doombot-exe: We want different denounce behavior for save vs submit. Submit is leading edge and save is trailing.

src/Assignments/ProblemIframe.tsx
doombot-exe: Oops, missed a cleanup item

src/Assignments/ProblemIframe.tsx
doombot-exe: I have corresponding changes to the backend behavior that hinges on a workbookID being included.

src/Assignments/ProblemIframe.tsx
doombot-exe: added down stream iterator comment

PanopticaRising (approved):

PR #165, opened 2020-09-23
Bugfix/red 137 iframe resize

red-tt-bear:


src/Assignments/ProblemIframe.tsx
```diff
@@ -139,39 +128,35 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
         setLoading(false);
     };
 
     return (
         <>
             { loading && <Spinner animation='border' role='status'><span className='sr-only'>Loading...</span></Spinner>}
             {error && <div>{error}</div>}
-            <iframe
+            <IframeResizer
+                // Using onInit instead of ref because:
```
PanopticaRising, line 143: Nice finds.

src/Assignments/ProblemIframe.tsx
```diff
@@ -139,39 +128,35 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+                // ref never get's set and a warning saying to use `forwardRef` comes up in the console
+                // Using forwardRef does not give you access to the iframe, rather it gives you access to 3 or 4 methods and properties (like `sendMessage`)
+                onInit={(iframe: IFrameComponent) => {
+                    // TODO do we need to unset the iframeref? As of right now it should not be required since it is always present within the component
+                    // If using dom elements the useRef is "Read Only", however I want control!
+                    (iframeRef as any).current = iframe;
```
PanopticaRising, line 149: I'd say, since we can hook-in via the event handlers, we don't need a ref for anything.

PanopticaRising (approved):

doombot-exe (approved):

src/Assignments/ProblemIframe.tsx
doombot-exe: upcoming changes for save and submit use the ref for grabbing the form element and GeoGebra-related scripts in the iframe

doombot-exe: have not tested the modified save/submit with these changes though...

PR #173, opened 2020-10-01
Feature/red 51 grant individual extensions

PanopticaRising:
I'll have updates to this, but please review in the meanwhile.


src/Courses/CourseInterfaces.tsx
```diff
@@ -50,9 +55,11 @@ export class UserObject {
     firstName?: string;
     lastName?: string;
     id: number = -1;
+    name: string;
 
     public constructor(init?:Partial<UserObject>) {
         Object.assign(this, init);
+        this.name = `${this.firstName} ${this.lastName}`;
```
red-tt-bear, line 62: Should we make this a getter instead? The code that calls it will remain the same but if the values update the getter will be correct

src/Courses/CourseInterfaces.tsx
```diff
@@ -96,10 +107,15 @@ export class NewCourseTopicObj extends TopicObject {
...
     deadDate: Date = new Date();
     partialExtend: boolean = false;
+    studentTopicOverride: any[] = [];
 
     public constructor(init?:Partial<NewCourseTopicObj>) {
-        super();
+        super(init);
         Object.assign(this, init);
+        
+        if (!_.isNull(init?.questions)) {
```
red-tt-bear, line 116: Why do we have to do this? is it for the default values?

src/Courses/CourseProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+import AxiosRequest from '../Hooks/AxiosRequest';
+import { CourseObject, UserObject } from './CourseInterfaces';
+
+interface CourseProviderProps {
+    children: React.ReactNode
+}
+
+const CourseContext = React.createContext<{
+        course: CourseObject, 
+        setCourse: React.Dispatch<React.SetStateAction<CourseObject>> | undefined,
```
red-tt-bear, line 14: Prefer question mark after variable name

src/Courses/CourseProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+        setCourse: undefined,
+        error: null,
+        users: [],
+        setUsers: undefined,
+    });
+
+export const useCourseContext = () => React.useContext(CourseContext);
+
+export const CourseProvider: React.FC<CourseProviderProps> = ({children}) => {
+    const { courseId } = useParams<{courseId: string | undefined}>();
```
red-tt-bear, line 29: In this case you can't use the question mark

src/Courses/CourseProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+    const [loading, setLoading] = useState<boolean>(false);
+    const [error, setError] = useState<string | null>(null);
+
+    useEffect(() => {
+        (async () => {
+            if (!courseId) return;
+            setLoading(true);
+            setError(null);
+            try {
+                const courseRespPromise = AxiosRequest.get(`/courses/${courseId}`);
```
red-tt-bear, line 41: Should put this in the backend api files

src/Courses/CourseProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+                const userRespPromise = getUsersForCourse({courseId: parseInt(courseId, 10)});
+
+                const [courseResp, userResp] = await Promise.all([courseRespPromise, userRespPromise]);
```
red-tt-bear, line 44: Nice!

src/Courses/CourseProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+                setCourse(new CourseObject(courseResp.data.data));
+
+                const usersArr = [];
+                for (let user of userResp.data.data) {
+                    usersArr.push(new UserObject(user));
+                }
+
+                setUsers(usersArr);
+            } catch (e) {
+                setError(e.response.data.message);
```
red-tt-bear, line 54: This is not safe, depending on how the request fails the whole page will crash, that's why it's better to do the type safe way and use e.message

src/Courses/CourseProvider.tsx
```diff
@@ -0,0 +1,71 @@
...
+                console.error(e);
```
red-tt-bear, line 55: We shouldn't console.error here, web requests can fail so this would not be a case where we would send the log to the server

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+};
+
+interface OverridesFormProps {
+    userId: number;
+    topic?: NewCourseTopicObj;
+    problem?: ProblemObject;
+}
+
+export const OverridesForm: React.FC<OverridesFormProps> = ({topic, userId, problem}) => {
+    const { register, handleSubmit, watch, errors, control, setValue } = useForm<Inputs>();
```
red-tt-bear, line 25: Nice!

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+                // Right now, we expect only one override per topic to be returned.
+                if (topicData.studentTopicOverride?.length === 1) {
+                    _.assign(topicData, topicData.studentTopicOverride[0]);
+                }
+
+                setValue('startDate', moment(topicData.startDate));
+                setValue('endDate', moment(topicData.endDate));
+                setValue('deadDate', moment(topicData.deadDate));
+            } catch (e) {
+                console.error(`Topic ${topic.id} or User ${userId} does not exist!`, e);
```
red-tt-bear, line 49: This shouldn't be a console.error it should be a setError (although it doesn't look like we have that)

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+            } finally {
+                setFormLoading(false);
```
red-tt-bear, line 51: I think I've just been putting this outside the try catch, in this case and all of my cases it didn't matter but this is the right way to do it (in case some error occurs in the catch block)

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+            }
+        })();
+    }, [topic, userId]);
```
red-tt-bear, line 54: I'm not sure the solution here but watching an entire object isn't great
If there is a local change it causes an api call to fire off, you really only care if the id changes
I understand why you did this, it's because the linter yells at you, we just need to figure it out at some point

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+                setFormLoading(false);
+            }
+        })();
+
+    }, [problem, userId]);
+
+    const updateTopic = async (courseTopicContentId: number, userId: number, extensions: {startDate: Moment, endDate: Moment, deadDate: Moment}) => {
+        try {
+            const res = await extendTopic({courseTopicContentId, userId, extensions});
+            console.log(res);
```
red-tt-bear, line 84: 👿

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+
+    const onSubmit = async (extensions: {startDate: Moment, endDate: Moment, deadDate: Moment} | {maxAttempts: number}) => {
+        setSubmitLoading(true);
+        console.log(extensions);
+        if (problem) {
+            await updateQuestions(problem.id, userId, extensions as {maxAttempts: number});
+        } else if (topic) {
+            await updateTopic(topic.id, userId, extensions as {startDate: Moment, endDate: Moment, deadDate: Moment});
+        } else {
+            console.error('Unhandled override case.');
```
red-tt-bear, line 109: Nice!

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+                <TextField inputRef={register} name="maxAttempts" defaultValue={question.maxAttempts} label='Max Attempts' />
+            </Grid>
+        </Grid>
+    );
+
+    const renderTopicOverrideForm = (topic: NewCourseTopicObj) => (
+        <>
+            <Grid item md={12}>
+                <MuiPickersUtilsProvider utils={MomentUtils}>
+                    <Controller
```
red-tt-bear, line 126: We've had issues with doing this with the `Controller`, is there any advantage
On the topic modal you couldn't control `Controller` with the value state
Unless there's a reason to we should keep it consistent

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+                </MuiPickersUtilsProvider>
+            </Grid>
+
+            <Grid item md={12}>
+                <MuiPickersUtilsProvider utils={MomentUtils}>
+                    <Controller
+                        as={<KeyboardDatePicker value="" onChange={() => {}} />}
+                        name="endDate"
+                        control={control}
+                        defaultValue={moment(topic.endDate)}
```
red-tt-bear, line 151: This should be value and not default value, otherwise when updates come into props it won't update (though I think the controller is a problem anyway)

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,238 @@
...
+    return (
+        <form onSubmit={handleSubmit(onSubmit)} style={{width: '100%', marginTop: '1.5rem'}}>
+            <Grid container justify='center'>
+                <Grid container item md={6} spacing={2}>
+                    {/* TODO: Use AnimatePresence for a better UX than the flicker or delay. */}
+                    {formLoading ? (
+                        <CircularProgress />
+                    ) : (
+                        <>
+                            {/* TODO: Ternary with problem overrides */}
```
red-tt-bear, line 200: Isn't that what you're doing?

src/Courses/Settings/SettingsPage.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+import './SettingsPage.css';
+
+interface SettingsPageProps {
+
+}
+
+export const SettingsPage: React.FC<SettingsPageProps> = ({}) => {
+    const {course, setCourse, users, setUsers} = useCourseContext();
+    const [selected, setSelected] = useState<{
```
red-tt-bear, line 17: no single lines...

src/Courses/Settings/SettingsPage.tsx
```diff
@@ -0,0 +1,49 @@
...
+            <Row>
+                <Col className='text-center'>
+                    <h1>Course/Individual Settings</h1>
+                </Col>
+            </Row>
+            <MaterialTriSelect course={course} users={users} selected={selected} setSelected={setSelected} />
+            <Drawer 
+                className='black-drawer'
+                anchor='bottom' 
+                open={!!(selected.user && selected.topic)}
```
red-tt-bear, line 32: Nit: Double bang or Boolean wrapper 🤔 I think we differ on opinion here but I think wrapping in Boolean() is easier to read

src/Courses/TopicsList.tsx
```diff
@@ -54,65 +56,119 @@ export const TopicsList: React.FC<TopicsListProps> = ({listOfTopics, flush, show
...
+        const now = moment();
+        if (_.isEmpty(topic.studentTopicOverride)) return [];
+
+        const activeExtensions: any[] = topic.studentTopicOverride.reduce((accum, extension) => {
+            if (now.isBetween(extension.startDate, extension.deadDate, 'day', '[]')) {
+                accum.push(extension);
+            }
+            return accum;
+        }, []);
+        console.log('Active Extensions: ', activeExtensions, topic.name);
```
red-tt-bear, line 70: You were doing so well commenting out the first one 🤣

src/Courses/TopicsList.tsx
```diff
@@ -54,65 +56,119 @@ export const TopicsList: React.FC<TopicsListProps> = ({listOfTopics, flush, show
...
-                            onChange={()=>{}}
-                            onAccept={(date: MaterialUiPickersDate) => {
-                                if (!date) return;
-                                updateTopicField(topic, 'startDate', date.toDate());
-                            }}
-                            inputVariant='outlined'
-                            disabled={userType === UserRole.STUDENT}
-                        />
+                        <MuiPickersUtilsProvider utils={MomentUtils}>
+                            <DateTimePicker
```
red-tt-bear, line 137: This is what I prefer over the controller above

red-tt-bear: When I first encountered the issue I had noticed we did it differently in two different places

red-tt-bear (changes requested):

src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
```diff
@@ -150,6 +163,19 @@ export const deleteTopic = async ({
     }
 };
 
+export const extendTopic = async ({
+    courseTopicContentId, userId,
+    extensions
+}: ExtendCourseTopicForUser): Promise<AxiosResponse<BackendAPIResponse>> => {
+    try {
+        return await AxiosRequest.put(
+            `${COURSE_TOPIC_PATH}/extend?courseTopicContentId=${courseTopicContentId}&userId=${userId}`, extensions
```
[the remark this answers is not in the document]
PanopticaRising: I considered that, but I thought doing that at the top added a `/` at the end.

src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
```diff
@@ -150,6 +163,19 @@ export const deleteTopic = async ({
     }
 };
 
+export const extendTopic = async ({
+    courseTopicContentId, userId,
```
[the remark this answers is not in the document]
PanopticaRising: I don't particularly see the benefit. If they were more verbose, or if we needed to add typings, sure, but I prefer to read horizontally than vertically.

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -40,19 +40,31 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
                     courseTopicContentId: parseInt(params.topicId, 10)
                 });
                 const problems: Array<ProblemObject> = res.data.data.questions;
+                
                 const topic = res.data.data.topic;
+                console.log(topic);
```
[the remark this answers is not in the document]
PanopticaRising: I told you this was just to preview!

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -40,19 +40,31 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
+                if (topic.studentTopicOverride?.length > 0) {
```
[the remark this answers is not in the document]
PanopticaRising: No, it's only greater than 0 if an override actually exists.

src/Components/MaterialTriSelect.tsx
```diff
@@ -0,0 +1,87 @@
...
+import { AnimatePresence, motion } from 'framer-motion';
+import _ from 'lodash';
+import React, { useEffect, useState } from 'react';
+import { CourseObject, UnitObject, UserObject, NewCourseTopicObj, ProblemObject, SettingsComponentType } from '../Courses/CourseInterfaces';
+import MultiSelectCardList from './MultiSelectCardList';
+
+interface MaterialTriSelectProps {
+    course: CourseObject;
+    users: UserObject[];
+    selected: {unit: UnitObject | undefined, topic: NewCourseTopicObj | undefined, problem: ProblemObject | undefined, user: UserObject | undefined};
```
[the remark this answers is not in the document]
PanopticaRising: You're right; sometimes I get confused about where I can use the ? and where I can't.

src/Components/MaterialTriSelect.tsx
```diff
@@ -0,0 +1,87 @@
+import { Grid } from '@material-ui/core';
+import { AnimatePresence, motion } from 'framer-motion';
+import _ from 'lodash';
+import React, { useEffect, useState } from 'react';
+import { CourseObject, UnitObject, UserObject, NewCourseTopicObj, ProblemObject, SettingsComponentType } from '../Courses/CourseInterfaces';
+import MultiSelectCardList from './MultiSelectCardList';
+
+interface MaterialTriSelectProps {
+    course: CourseObject;
```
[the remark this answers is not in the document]
PanopticaRising: I figure we might reuse it for other things, so I threw it here too. It really is just a selectable list, so it should be reusable with some cleanup.

src/Courses/CourseDetailsPage.tsx
```diff
@@ -76,13 +58,13 @@ export const CourseDetailsPage: React.FC<CourseDetailsPageProps> = () => {
                         setStudentNameAndId(null);
                     }}>
                     <Tab eventKey={CourseDetailsTabs.DETAILS} title={CourseDetailsTabs.DETAILS}  style={{marginBottom:'10px'}}>
-                        <CourseDetailsTab course={course} error={error} loading={loading} setCourse={setCourse} />
+                        <CourseDetailsTab course={course} error={error} loading={false} setCourse={setCourse} />
```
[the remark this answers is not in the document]
PanopticaRising: No, this was related to loading being raised up to the wrapper component that provides the backdrop. I'll add a TODO to clean it up; we talked about having subcomponents queue up loading times into that backdrop.

src/Courses/CourseDetailsTabs/CourseDetailsTab.tsx
```diff
@@ -14,7 +14,7 @@ interface CourseDetailsTabProps {
     course?: CourseObject;
     loading: boolean;
     error: string | null;
-    setCourse?: (course: CourseObject) => void;
+    setCourse?: React.Dispatch<React.SetStateAction<CourseObject>>;
```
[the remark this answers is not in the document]
PanopticaRising: This is the type for the function that `useState` returns.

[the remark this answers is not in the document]
PanopticaRising: image

src/Courses/CourseDetailsTabs/ActiveTopics.tsx
```diff
@@ -2,19 +2,38 @@ import React, { useState, useEffect } from 'react';
...
+    const userId: number = getUserId();
 
     useEffect(()=>{
         (async () => {
             const res = await AxiosRequest.get(`/courses/topics?isOpen=true&courseId=${course.id}`);
             console.log(res.data.data);
-            const activeTopics = res.data?.data?.map((topic: any) => new NewCourseTopicObj(topic));
+            let activeTopics = res.data?.data?.map((topic: any) => new NewCourseTopicObj(topic));
+
+            // Currently, the backend returns topics that have extensions for any users.
```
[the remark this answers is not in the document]
doombot-exe: I would think this continues - unless we have good reason to split off a separate endpoint on the backend for this specific purpose. My assumption is that the individual overrides content wants to know about all overrides so as to differentiate the appearance of cards corresponding to existing overrides.

src/Courses/CourseInterfaces.tsx
PanopticaRising: This applies the typing/constructor logic to all the subobjects that come back from the API.

src/Courses/CourseProvider.tsx
PanopticaRising: Actually, I think you can use the question mark because it's in an object.

src/Courses/CourseProvider.tsx
PanopticaRising: I'm going to push this off for another cleanup.

src/Courses/Settings/OverridesForm.tsx
PanopticaRising: This is supposed to be an impossible case, so it seems like something we'd still want to log.

src/Courses/Settings/OverridesForm.tsx
PanopticaRising: A local change shouldn't matter unless it overwrites the object and it gets a new memory address.

src/Courses/Settings/SettingsPage.tsx
PanopticaRising: We do differ :joy:

src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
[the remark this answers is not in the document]
red-tt-bear: Having a slash at the end and adding a ? afterwards should still work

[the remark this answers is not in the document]
red-tt-bear: Regardless we're thinking about changing this layer in the future so I don't want to hold anything off, this is good enough for now

src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
```diff
@@ -258,6 +299,25 @@ export const getQuestions = async ({
...
 };
 
+export const extendQuestion = async ({
+    courseTopicQuestionId, 
+    userId,
+    extensions
+}: ExtendCourseTopicQuestionsForUser): Promise<AxiosResponse<BackendAPIResponse>> => {
+    try {
+        return await AxiosRequest.put(
+            `${COURSE_QUESTION_PATH}/extend`, extensions, {
```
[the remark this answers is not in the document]
red-tt-bear: We can leave this for now

src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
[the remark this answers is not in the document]
red-tt-bear: I like vertically because you can comment out easier, but it doesn't matter

src/Components/MaterialTriSelect.tsx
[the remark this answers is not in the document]
red-tt-bear: We'll make generic in the future

src/Courses/CourseDetailsTabs/ActiveTopics.tsx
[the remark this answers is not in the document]
red-tt-bear: @doombot-exe 
It wouldn't be a separate endpoint
The get active topics endpoint we're using should have get the role from the session
If professor get all topics where the date is between: (earliest start override or topic start date) and (latest dead override dead date or topic dead date)
If student use your specific override

so the professor sees all active topics
student's see all their active topics
both with overrides

src/Courses/CourseInterfaces.tsx
red-tt-bear: Gotcha you mean for instance of, I wonder if there is a more efficient way of assigning a prototype 🤔

src/Courses/Settings/OverridesForm.tsx
red-tt-bear: True, but even if that's the case if the id doesn't change I don't want to refetch

src/Courses/Settings/OverridesForm.tsx
red-tt-bear: Talked about it in scrum and @PanopticaRising showed me that it works as expected, might have been how we did the `as` in the past

src/Courses/CourseDetailsTabs/ActiveTopics.tsx
```diff
@@ -2,19 +2,38 @@ import React, { useState, useEffect } from 'react';
...
             console.log(res.data.data);
-            const activeTopics = res.data?.data?.map((topic: any) => new NewCourseTopicObj(topic));
+            let activeTopics = res.data?.data?.map((topic: any) => new NewCourseTopicObj(topic));
+
+            // TODO: Currently, the backend returns topics that have extensions for any users.
+            // This filters out those topics and hides them.
+            if (userType === UserRole.STUDENT) {
+                activeTopics = activeTopics.filter((topic: NewCourseTopicObj) => {
+                    // If this actually is an active topic, keep it.
+                    if (moment().isBetween(topic.startDate, topic.deadDate, 'day', '[]')) {
```
red-tt-bear, line 29: What is with the extra parameters here?

src/Courses/CourseDetailsTabs/ActiveTopics.tsx
```diff
@@ -2,19 +2,38 @@ import React, { useState, useEffect } from 'react';
...
+    const userId: number = getUserId();
 
     useEffect(()=>{
         (async () => {
             const res = await AxiosRequest.get(`/courses/topics?isOpen=true&courseId=${course.id}`);
             console.log(res.data.data);
-            const activeTopics = res.data?.data?.map((topic: any) => new NewCourseTopicObj(topic));
+            let activeTopics = res.data?.data?.map((topic: any) => new NewCourseTopicObj(topic));
+
+            // TODO: Currently, the backend returns topics that have extensions for any users.
```
red-tt-bear, line 24: This filter is no longer used, we fixed the backend, TODO take it out

src/Courses/CourseDetailsTabs/EmailComponentWrapper.tsx
```diff
@@ -99,6 +103,13 @@ export const EmailComponentWrapper: React.FC<EmailComponentWrapperProps> = ({ us
                             onClick: (event: any, user: any) => setShowConfirmDelete({state: true, user}),
                             position: 'row'
                         },
+                        {
+                            // eslint-disable-next-line react/display-name
```
red-tt-bear, line 107: Why are we disabling this, we should add a comment

src/Courses/CourseDetailsTabs/EmailComponentWrapper.tsx
```diff
@@ -99,6 +103,13 @@ export const EmailComponentWrapper: React.FC<EmailComponentWrapperProps> = ({ us
...
+                            icon: () => <Link to={(loc: any) => ({...loc, pathname: `${loc.pathname}/settings`})}><MdLaunch style={{color: 'black'}} /></Link>,
+                            tooltip: 'Go to Extensions',
+                            onClick: (event: any, user: any) => null,
```
red-tt-bear, line 110: We should avoid any

src/Courses/CourseDetailsTabs/EmailComponentWrapper.tsx
red-tt-bear: Nit ^

src/Courses/CourseDetailsTabs/EmailComponentWrapper.tsx
red-tt-bear: Nit ^

src/Courses/Settings/OverridesForm.tsx
```diff
@@ -0,0 +1,279 @@
...
+                setFormLoading(false);
+            }
+        })();
+
+    }, [problem, userId]);
+
+    const updateTopic = async (courseTopicContentId: number, userId: number, extensions: {startDate: Moment, endDate: Moment, deadDate: Moment}) => {
+        try {
+            const res = await extendTopic({courseTopicContentId, userId, extensions});
+            console.log(res);
```
red-tt-bear, line 100: You resolved the conversation without removing the console.log....

red-tt-bear (approved):

package.json
```diff
@@ -79,6 +82,6 @@
     "eslint-plugin-react": "^7.19.0",
     "http-proxy-middleware": "^1.0.5",
     "react-axe": "^3.4.1",
-    "react-hook-form-devtools": "^1.2.0"
```
red-tt-bear, line 82: The new version of this doesn't seem to exist in npm
When I put it back to the old version it updated package-lock.json as it hadn't been there....
@PanopticaRising can you please let me know what the correct action is?

red-tt-bear, line 82: This isn't in the existing package-lock.json
The new version listed doesn't exist in npm
@PanopticaRising please clarify what is correct

PR #179, opened 2020-10-08
Bugfix/extension bugs

PanopticaRising:


src/Courses/Settings/OverridesForm.tsx
```diff
@@ -36,27 +36,22 @@ export const OverridesForm: React.FC<OverridesFormProps> = ({topic, userId, prob
...
         setDefaultProblem(problem);
-        setIsSubmitSuccesful(false);
         reset();
     }, [topic, problem, userId]);
 
     // Get Topic Override information
     useEffect(()=>{
         if (!topic || problem !== undefined) return;
         setFormLoading(true);
-        setSubmitError('');
```
red-tt-bear, line 59: Are we not clearing the error?
Does the error clear when you move to another override?

red-tt-bear (approved):

src/Courses/Settings/OverridesForm.tsx
PanopticaRising: Yes, the error clears when you move to another override. The logic has been passed to react-hook-forms rather than managing our own state.

PR #182, opened 2020-10-09
preventDefault all 'submit' and check clicked btn

doombot-exe:
submission now blocked for 4 seconds following any successful submit
'submitted at:' is only updated on 'submitAnswers' click (in case of other buttons)

### DIFF
--- a/src/Assignments/ProblemIframe.tsx
+++ b/src/Assignments/ProblemIframe.tsx
@@ -141,7 +141,6 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
     function insertListener(problemForm: HTMLFormElement) {
         problemForm.addEventListener('submit', (e) => { e.preventDefault();});
         problemForm.addEventListener('submit', _.debounce(() => {
-            // event.preventDefault();
             if (_.isNil(problemForm)) {
                 console.error('Hijacker: Could not find the form when submitting the form');
                 setError('An error occurred');
--- a/src/Assignments/ProblemIframe.tsx
+++ b/src/Assignments/ProblemIframe.tsx
@@ -139,8 +139,7 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
     }
 
     function insertListener(problemForm: HTMLFormElement) {
-        problemForm.addEventListener('submit', (e) => { e.preventDefault();});
-        problemForm.addEventListener('submit', _.debounce(() => {
+        const submitHandler = _.debounce( () => {
             if (_.isNil(problemForm)) {
                 console.error('Hijacker: Could not find the form when submitting the form');
                 setError('An error occurred');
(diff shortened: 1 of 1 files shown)


src/Assignments/ProblemIframe.tsx
```diff
@@ -137,8 +139,8 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
     }
 
     function insertListener(problemForm: HTMLFormElement) {
-        problemForm.addEventListener('submit', _.debounce((event: { preventDefault: () => void; }) => {
-            event.preventDefault();
+        problemForm.addEventListener('submit', (e) => { e.preventDefault();});
+        problemForm.addEventListener('submit', _.debounce(() => {
```
PanopticaRising, line 143: I'm not sure about the behavior of `preventDefault` across multiple listeners. I think I'd prefer that we combine the two calls into one as you had before, with the debounce logic _after_ the `preventDefault()`.

PanopticaRising (approved):

src/Assignments/ProblemIframe.tsx
doombot-exe: updated

src/Assignments/ProblemIframe.tsx
```diff
@@ -137,21 +139,31 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
             }
             prepareAndSubmit(problemForm, clickedButton);
-        }, 4000, {leading: true, trailing:false}));
+        };
+        const debouncedSubmitHandler = _.debounce(submitHandler, 4000, { leading: true, trailing: false });
+
+        problemForm.addEventListener('submit', (event: { preventDefault: () => void; }) => {
+            event.preventDefault();
+            const clickedButton = problemForm.querySelector('.btn-clicked') as HTMLButtonElement;
+            const shouldSubmit = (clickedButton.name === 'submitAnswers');
```
red-tt-bear, line 160: Don't need parenthesis here

red-tt-bear: Nit ^

src/Assignments/ProblemIframe.tsx
```diff
@@ -137,21 +139,31 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
-            let clickedButton = problemForm.querySelector('.btn-clicked') as HTMLButtonElement;
             if (_.isNil(clickedButton)) {
                 setError('Hijacker: An error occurred');
                 console.error('Could not find the button that submitted the form');
                 return;
             }
             prepareAndSubmit(problemForm, clickedButton);
-        }, 4000, {leading: true, trailing:false}));
+        };
+        const debouncedSubmitHandler = _.debounce(submitHandler, 4000, { leading: true, trailing: false });
```
red-tt-bear, line 155: I wonder what happens if you set leading and trailing to true 🤔 

I would think it would run immediately and if you clicked again it would run after the delay, multiple clicks would dealy the dealy event

red-tt-bear (approved):

PR #186, opened 2020-10-09
Bugfix/red 157 debounce submissions

doombot-exe:
stash form contents on submission, and whenever form contents are the same as what was submitted:
 - disable submit button
 - change the button text to "Submitted" (stashing the old text)

Whenever contents change (during save contents call), redo comparison and toggle based on equivalence.

### DIFF
--- a/src/Assignments/ProblemIframe.tsx
+++ b/src/Assignments/ProblemIframe.tsx
@@ -70,30 +70,29 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
         setLastSubmittedAt?.(null);
         setLastSavedAt?.(null);
         setLastSubmission({});
-        updateSubmitActive();
     }, [problem.id]);
 
     const updateSubmitActive = _.throttle(() => {
         const submitButtons = iframeRef.current?.contentWindow?.document.getElementsByName('submitAnswers') as NodeListOf<HTMLButtonElement>;
         const problemForm = iframeRef.current?.contentWindow?.document.getElementById('problemMainForm') as HTMLFormElement;
-        // shouldn't happen - called only onLoad or after interaction with already loaded srcdoc
-        // console.error()?
-        if (_.isNil(submitButtons) || _.isNil(problemForm)) { return; }
+        // called only onLoad or after interaction with already loaded srcdoc - so form will exist, unless bad problemPath
+        // no console.error because exam problems (and static problems) will not have 'submitAnswers'
+        if (_.isNil(submitButtons) || _.isNil(problemForm)) {return;}
 
         const isClean = _.isEqual(formDataToObject(new FormData(problemForm)), lastSubmission);
 
         submitButtons.forEach((button: HTMLButtonElement) => {
             if (isClean) {
-                button.disabled = true;
+                button.setAttribute('disabled','true');
                 // invisibly stash the button's label (in case there are multiple submit buttons)
-                button.textContent = button.value;
-                button.value = 'Submitted';
+                button.setAttribute('textContent', button.value);
+                button.setAttribute('value', 'Submitted');
             } else {
                 button.removeAttribute('disabled');
-                if (!_.isNil(button.textContent) && button.textContent != '') {
+                if (button.textContent) {
                     // put it back and clear the stash - just in case
-                    button.value = button.textContent;
-                    button.textContent = '';
+                    button.setAttribute('value', button.textContent);
+                    button.removeAttribute('textContent');
                 }
             }
         });
@@ -123,7 +122,8 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
         let formData = new FormData(problemForm);
         if (!_.isNil(clickedButton)) {
             // current state will never match submission unless we save it before including clickedButton
-            setLastSubmission(formDataToObject(formData)); 
+            // but we only want to save the current state if the button was 'submitAnswers'
+            const saveMeLater = formDataToObject(formData); 
             formData.set(clickedButton.name, clickedButton.value);
             try {
                 const result = await postQuestionSubmission({
@@ -138,6 +138,7 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
                 setRenderedHTML(result.data.data.rendererData.renderedHTML);
                 if (clickedButton.name === 'submitAnswers'){
                     setProblemStudentGrade(result.data.data.studentGrade);
+                    setLastSubmission(saveMeLater);
                     setLastSubmittedAt?.(moment());
                 }
             } catch (e) {
@@ -171,21 +172,14 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
     }
 
     function insertListeners(problemForm: HTMLFormElement) {
-        const throttledSubmitHandler = _.throttle(prepareAndSubmit, 2000, { leading: true, trailing: true });
         const debouncedSubmitHandler = _.debounce(prepareAndSubmit, 2000, { leading: false, trailing: true });
 
         // submission of problems will trigger updateSubmitActive @onLoad
         // because re-submission of identical answers is blocked, we expect srcdoc to change
-        // even if our assumption fails, 'actual' submission is throttled
         problemForm.addEventListener('submit', (event: { preventDefault: () => void; }) => {
             event.preventDefault();
             const clickedButton = problemForm.querySelector('.btn-clicked') as HTMLButtonElement;
-            if (clickedButton.name === 'submitAnswers') {
-                // submission is blocked until contents change - throttle will suffice
-                throttledSubmitHandler(problemForm, clickedButton);
-            } else {
-                prepareAndSubmit(problemForm, clickedButton);
-            }
+            prepareAndSubmit(problemForm, clickedButton);
         });
 
         problemForm.addEventListener('input', () => {
@@ -213,7 +207,7 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
             // check that the submit url is accurate
             const submitUrl = problemForm.getAttribute('action');
             const checkId = submitUrl?.match(/\/backend-api\/courses\/question\/([0-9]+)\?/);
-            if (checkId && parseInt(checkId[1],10) != problem.id) {
+            if (checkId && parseInt(checkId[1],10) !== problem.id) {
                 console.error('Something went wrong. This problem is reporting an ID that is incorrect');
                 setError('This problem ID is out of sync.');
                 return;
--- a/src/Assignments/ProblemIframe.tsx
+++ b/src/Assignments/ProblemIframe.tsx
@@ -34,7 +34,9 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
     const [renderedHTML, setRenderedHTML] = useState<string>('');
     const [loading, setLoading] = useState(true);
     const [error, setError] = useState('');
+    const [lastSubmission, setLastSubmission] = useState({});
     const height = '100vh';
+    let isClean = false;
 
     const { setLastSavedAt, setLastSubmittedAt } = useCurrentProblemState();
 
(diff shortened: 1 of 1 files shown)


src/Assignments/ProblemIframe.tsx
```diff
@@ -34,7 +34,9 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
     const [renderedHTML, setRenderedHTML] = useState<string>('');
     const [loading, setLoading] = useState(true);
     const [error, setError] = useState('');
+    const [lastSubmission, setLastSubmission] = useState({});
     const height = '100vh';
+    let isClean = false;
```
red-tt-bear, line 39: This should be scoped to the function since it is only used in there

src/Assignments/ProblemIframe.tsx
```diff
@@ -107,6 +134,7 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
                 if (clickedButton.name === 'submitAnswers'){
                     setProblemStudentGrade(result.data.data.studentGrade);
                     setLastSubmittedAt?.(moment());
+                    updateSubmitActive(true);
```
red-tt-bear, line 137: Shouldn't this be false?
Once submit is clicked it should be disabled?

src/Assignments/ProblemIframe.tsx
```diff
@@ -119,7 +147,8 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
                 // TODO: impossi-log console.error()
                 setError(`No grades id for problem #${problem.id}`);
                 return;
-            }    
+            }
+            updateSubmitActive(_.isEqual(formDataToObject(formData), lastSubmission));
```
red-tt-bear, line 151: Since it's part of save isn't there a 2 second delay?

src/Assignments/ProblemIframe.tsx
```diff
@@ -201,6 +229,8 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
                 return;
             }
             insertListener(problemForm);
+            let formData = new FormData(problemForm);
+            updateSubmitActive(_.isEqual(formDataToObject(formData), lastSubmission));
```
red-tt-bear, line 233: This is in the load handler, isn't this always true?

src/Assignments/ProblemIframe.tsx
doombot-exe: isClean = true, so disabled, yes. Sorry about the nomenclature, I was trying to do everything based on isClean - but now it's just a parameter that is passed to updateSubmitActive -- think of it as updateSubmitActive(isClean = true)

src/Assignments/ProblemIframe.tsx
```diff
@@ -67,8 +69,32 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
         // when problem changes, reset lastsubmitted and lastsaved
         setLastSubmittedAt?.(null);
         setLastSavedAt?.(null);
+        setLastSubmission({});
+        updateSubmitActive(false);
     }, [problem.id]);
 
+    const updateSubmitActive = (incomingIsClean: boolean | undefined) => {
```
red-tt-bear, line 76: Method name needs to be renamed
updateSubmitActive true seems like you are saying you want to make submit active
However the parameter is the negative meaning (which you don't see without a good idea)

a clean object should be disabled

src/Assignments/ProblemIframe.tsx
doombot-exe: yes, after updating, the button will re-activate once the saveState is updated

src/Assignments/ProblemIframe.tsx
doombot-exe: this is var vs let?

src/Assignments/ProblemIframe.tsx
red-tt-bear: On submit calls onLoad so this does need to be here

src/Assignments/ProblemIframe.tsx
red-tt-bear: Let's add throttle

src/Assignments/ProblemIframe.tsx
red-tt-bear: Let's remove

src/Assignments/ProblemIframe.tsx
red-tt-bear: Let's change to isActive

red-tt-bear: remove "| undefined"

src/Assignments/ProblemIframe.tsx
red-tt-bear: Alright let's rename

red-tt-bear: You're mistake was not using the dirty bit!

src/Assignments/ProblemIframe.tsx
```diff
@@ -36,18 +36,18 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
 
     useEffect(()=>{
-        setLoading(true);
         // We need to reset the error state since a new call means no error
         setError('');
         // If you don't reset the rendered html you won't get the load event
         // Thus if you go to an error state and back to the success state
         // The rendered html will never call load handler which will never stop loading
         setRenderedHTML('');
+        // srcdoc='' triggers onLoad with setLoading(false) so setLoading(true) isn't effective until now
```
red-tt-bear, line 49: srcdoc should trigger onLoad regardless, I'm not sure of the benefit of moving this down but if it works it works, we can talk about it later

src/Assignments/ProblemIframe.tsx
```diff
@@ -67,8 +69,36 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+            } else {
+                button.removeAttribute('disabled');
+                if (!_.isNil(button.textContent) && button.textContent != '') {
+                    // put it back and clear the stash - just in case
+                    button.value = button.textContent;
+                    button.textContent = '';
+                }
+            }
+        });
+    }, 1000, {leading:true, trailing:true});
```
red-tt-bear, line 100: 1 second seems a bit long, I would think it should be something more like 100ms to 500ms

src/Assignments/ProblemIframe.tsx
doombot-exe: right, when srcdoc = '' triggers onLoad, then onLoad does setLoading(false), and then the renderedHTML gets updated with the actual problem contents, meanwhile setLoading remains false - which is why we see that the GeoGebra problems aren't being hidden while they load.

red-tt-bear: Great catch!!!

src/Assignments/ProblemIframe.tsx
```diff
@@ -67,8 +69,36 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
         setLastSavedAt?.(null);
+        setLastSubmission({});
+        updateSubmitActive();
     }, [problem.id]);
 
+    const updateSubmitActive = _.throttle(() => {
+        const submitButtons = iframeRef.current?.contentWindow?.document.getElementsByName('submitAnswers') as NodeListOf<HTMLButtonElement>;
+        const problemForm = iframeRef.current?.contentWindow?.document.getElementById('problemMainForm') as HTMLFormElement;
+        // shouldn't happen - called only onLoad or after interaction with already loaded srcdoc
+        // console.error()?
```
red-tt-bear, line 80: Yeah we should have this on all impossible cases

src/Assignments/ProblemIframe.tsx
```diff
@@ -67,8 +69,36 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+        if (_.isNil(submitButtons) || _.isNil(problemForm)) { return; }
+
+        const isClean = _.isEqual(formDataToObject(new FormData(problemForm)), lastSubmission);
+
+        submitButtons.forEach((button: HTMLButtonElement) => {
+            if (isClean) {
+                button.disabled = true;
```
red-tt-bear, line 87: Odd way of interacting with dom, these should be done with setAttribute

src/Assignments/ProblemIframe.tsx
```diff
@@ -67,8 +69,36 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+                // invisibly stash the button's label (in case there are multiple submit buttons)
+                button.textContent = button.value;
+                button.value = 'Submitted';
+            } else {
+                button.removeAttribute('disabled');
+                if (!_.isNil(button.textContent) && button.textContent != '') {
```
red-tt-bear, line 93: !== please

src/Assignments/ProblemIframe.tsx
```diff
@@ -67,8 +69,36 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+                    // put it back and clear the stash - just in case
+                    button.value = button.textContent;
+                    button.textContent = '';
```
red-tt-bear, line 96: setAttribute please

src/Assignments/ProblemIframe.tsx
red-tt-bear: you could check for falsey value: `if (button.textContent) {}`

src/Assignments/ProblemIframe.tsx
```diff
@@ -92,6 +122,8 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
 
         let formData = new FormData(problemForm);
         if (!_.isNil(clickedButton)) {
+            // current state will never match submission unless we save it before including clickedButton
+            setLastSubmission(formDataToObject(formData));
```
red-tt-bear, line 126: This will occur on all button clicks including `return to default`, we should explicitly check for submit button

src/Assignments/ProblemIframe.tsx
```diff
@@ -138,41 +170,30 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+        // even if our assumption fails, 'actual' submission is throttled
         problemForm.addEventListener('submit', (event: { preventDefault: () => void; }) => {
             event.preventDefault();
             const clickedButton = problemForm.querySelector('.btn-clicked') as HTMLButtonElement;
-            const shouldSubmit = (clickedButton.name === 'submitAnswers');
-            if (shouldSubmit) {
-                debouncedSubmitHandler(clickedButton);
+            if (clickedButton.name === 'submitAnswers') {
+                // submission is blocked until contents change - throttle will suffice
+                throttledSubmitHandler(problemForm, clickedButton);
```
red-tt-bear, line 185: Why do we need this throttled, with these changes the user can't spam the submit button

red-tt-bear: let's get rid of if else

src/Assignments/ProblemIframe.tsx
```diff
@@ -67,8 +69,35 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
         setLastSavedAt?.(null);
+        setLastSubmission({});
     }, [problem.id]);
 
+    const updateSubmitActive = _.throttle(() => {
+        const submitButtons = iframeRef.current?.contentWindow?.document.getElementsByName('submitAnswers') as NodeListOf<HTMLButtonElement>;
+        const problemForm = iframeRef.current?.contentWindow?.document.getElementById('problemMainForm') as HTMLFormElement;
+        // called only onLoad or after interaction with already loaded srcdoc - so form will exist, unless bad problemPath
+        // no console.error because exam problems (and static problems) will not have 'submitAnswers'
+        if (_.isNil(submitButtons) || _.isNil(problemForm)) {return;}
```
red-tt-bear, line 80: FYI you don't need the brackets

red-tt-bear: I personally always like having the brackets but I also put it on a newline, @PanopticaRising would disagree with me though I think

src/Assignments/ProblemIframe.tsx
```diff
@@ -67,8 +69,35 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+            if (isClean) {
+                button.setAttribute('disabled','true');
+                // invisibly stash the button's label (in case there are multiple submit buttons)
+                button.setAttribute('textContent', button.value);
+                button.setAttribute('value', 'Submitted');
+            } else {
+                button.removeAttribute('disabled');
+                if (button.textContent) {
+                    // put it back and clear the stash - just in case
+                    button.setAttribute('value', button.textContent);
```
red-tt-bear, line 94: NIT: probably should be button.getAttribute

red-tt-bear (approved):

src/Assignments/ProblemIframe.tsx
doombot-exe: getAttribute causes complaint with only the false-y check from the if statement.

src/Assignments/ProblemIframe.tsx
doombot-exe: looks like I'll split the difference

PR #195, opened 2020-10-20
Feature/red 119 limit random seeds

PanopticaRising:


.eslintrc.json
```diff
@@ -45,6 +45,7 @@
         ],
         "react/prop-types": [0, {}],
         "no-unused-vars": "off",
-        "@typescript-eslint/no-unused-vars": "warn"
+        "@typescript-eslint/no-unused-vars": "warn",
+        "no-console": "warn",
```
red-tt-bear, line 49: Trailing comma causes errors in json files (js files are fine)
Different apps seem to handle it differently, when there was an extra comma in tsconfig.json I got some super specific non related error at some point in the future

red-tt-bear: Also nice!

src/Courses/TopicSettings/GenericFormInputs.tsx
```diff
@@ -1,133 +1,247 @@
...
+            <ErrorMessage name={name} errors={errors} />
+        </>
+    );
+};
+
+const examProblemFieldNamePrefix = 'courseQuestionAssessmentInfo';
+
+export const RandomSeedSet: React.FC<{}> = () => {
+    const { control, errors } = useFormContext();
+    const name = `${examProblemFieldNamePrefix}.randomSeedSet`;
```
red-tt-bear, line 100: what if different object structures were using the same form input? This seems like we are breaking down the ui for specific objects

Also if we are going to do this we should add some type checking:
```
const name = `${examProblemFieldNamePrefix}.${nameof<CourseQuestionAssessmentInfo>('randomSeedSet')}`;
```

Right now the rename to version was done on the backend and in these form elements that use strings but not in the model and we're not catching that

src/Courses/TopicSettings/GenericFormInputs.tsx
```diff
@@ -1,133 +1,247 @@
...
+
+export const MultipleProblemPaths: React.FC = ({}) => {
+    const fieldArrayName = 'additionalProblemPaths';
+    const nestedName = `${examProblemFieldNamePrefix}.${fieldArrayName}`
+    const { register, errors } = useFormContext();
+    const { fields, append, remove } = useFieldArray({
+        name: nestedName
+    });
+    
+    console.log(nestedName, 'MultiplePaths Fields', fields);
```
red-tt-bear, line 141: Your new eslint rule let one in 😈

src/Courses/TopicSettings/GenericFormInputs.tsx
PanopticaRising: This seems like a good candidate to address in a new PR.

src/Courses/TopicSettings/ProblemSettings.tsx
```diff
@@ -96,59 +143,78 @@ export const ProblemSettings: React.FC<ProblemSettingsProps> = ({selected, setSe
         }
     };
 
+    const rendererPreview = () => {
+        return (
+            <ProblemIframe 
+                problem={new ProblemObject({  })} 
+                setProblemStudentGrade={() => { }} 
+                workbookId={0}
```
red-tt-bear, line 151: Why are we hardcoding a workbookId?

src/Courses/TopicSettings/ProblemSettings.tsx
```diff
@@ -96,59 +143,78 @@ export const ProblemSettings: React.FC<ProblemSettingsProps> = ({selected, setSe
...
-                        {OptionalField(control)}
+                    {(updateAlertMsg !== '') && <Grid md={12} item><Alert variant={updateAlertType}>{updateAlertMsg}</Alert></Grid>}
+                    <Grid container item md={12} spacing={3}>
+                        <Grid item container md={12}><h1>Problem Settings</h1></Grid>
+                        <Grid item md={8}>
+                    Enter the path to the problem on the Rederly server. This is prefaced either 
+                    with <code>Library/</code> or <code>Contrib/</code> if this problem is included 
+                    in the <Link to='https://github.com/openwebwork/webwork-open-problem-library'>OPL</Link> or <code>private/</code> if 
+                    this problem has been uploaded to your private Rederly folder.
+                            {topic.topicTypeId === 2 ?
```
red-tt-bear, line 169: NIT: We really need to add more enums

src/Courses/TopicSettings/ProblemSettings.tsx
red-tt-bear: Not called, @PanopticaRising will remove

red-tt-bear (approved):

PR #197, opened 2020-10-21
Feature/red 144 exam submit

doombot-exe:
Supports assessments:
* start new version
* submit save-state of problems
* end exam early
* show/hide problems after exam closes based on config
* hot modal-on-modal action


src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
```diff
@@ -248,6 +251,24 @@ export const putQuestionGrade = async ({
     }
 };
 
+export const putQuestionGradeInstance = async ({
+    id,
+    data
+}: PutQuestionGradeInstanceOptions): Promise<AxiosResponse<PutQuestionGradeInstanceResponse>> => {
+    try {
+        console.log('putting question grade instance');
```
red-tt-bear, line 259: 😈 No console.logs

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -6,98 +6,305 @@ import ProblemIframe from './ProblemIframe';
...
+            show: true,
+            headerContent: <h5>Submit my exam</h5>,
+            bodyContent: 'You are about to use one of your graded submissions. Proceed?',
+            onConfirm: async () => await getResultsOfSubmission(topicId, versionId),
+            onHide: () => setConfirmationParameters(DEFAULT_CONFIRMATION_PARAMETERS),
+        });
+    };
+
+    const confirmStartNewVersion = (topic: TopicObject, actualVersionsRemaining?: number, message?: string) => {
+        actualVersionsRemaining = actualVersionsRemaining ?? versionsRemaining; // retrieve from state if not supplied
```
red-tt-bear, line 141: When would it not be supplied?
Is state accurate?
If state is not accurate what is the side effect?

src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
doombot-exe: yeah, I forgot to cleanup 😝

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -6,98 +6,305 @@ import ProblemIframe from './ProblemIframe';
...
+            </div>,
+            onHide: () => setConfirmationParameters(DEFAULT_CONFIRMATION_PARAMETERS),
+            onConfirm: async () => await loadNewVersion(topic)
+        });
+    };
+
+    const confirmEndVersion = () => {
+        let message = '';
+        if (!_.isNil(attemptsRemaining) && attemptsRemaining > 0) {
+            const nit = (attemptsRemaining === 1) ? 'attempt' : 'attempts';
```
red-tt-bear: 🤣 nice

You could just do it with the s itself, you don't have to do the whole word, either way, is fine it doesn't matter
This way is probably more readable

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -6,98 +6,305 @@ import ProblemIframe from './ProblemIframe';
...
+            // student had no attempts remaining, so the backend already knows the version is closed
+            setConfirmationParameters(DEFAULT_CONFIRMATION_PARAMETERS);
+            setModalLoading(false);
+        }
+    };
+
+    const generateScoreTable = (data: any) => {
+        const { problemScores, bestVersionScore, bestOverallVersion } = data;
+        return (
+            <div className="d-flex flex-column">
```
red-tt-bear, line 268: Absolutely fine for now, however, I think this is one of those cases where we have the code styled to the developer, we've been using a lot of material UI grids, I wonder if this should be refactored to be consistent in the future

@PanopticaRising thoughts?

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -114,25 +321,76 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
         return <div>{error}</div>;
     }
 
-    if (_.isEmpty(problems)) return <div>There was an error loading this assignment.</div>;
+    // if (_.isEmpty(problems)) return <div>There was an error loading this assignment.</div>;
```
red-tt-bear, line 324: Not a bad point, an empty array does not necessarily mean there is an error, however, I wonder how this impacts the rest of the application

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -114,25 +321,76 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
 
-    if (problems === null || selectedProblemId === null) return <div>Loading...</div>;
+    if (problems === null || selectedProblemId === null) return (
```
red-tt-bear, line 326: Our code styling has become inconsistent between the three of us, we should prioritize prettier sooner or later to make sure our code looks the same everywhere

src/Components/ModalWithChildren.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { Modal } from 'react-bootstrap';
+
+interface loginProps {
```
red-tt-bear: Copy pasta issue!

Should be named ModalWithChildrenProps

src/Components/ModalWithChildren.tsx
```diff
@@ -0,0 +1,36 @@
...
+    show: boolean;
+    header: string | JSX.Element;
+    children: React.ReactNode;
+    onClose?: () => unknown;
+}
+
+/**
+ * This component renders a button, and displays a Modal to log in with it.
```
red-tt-bear: Copy Pasta: This comment does not belong here

src/Courses/CourseInterfaces.tsx
```diff
@@ -162,15 +182,36 @@ export class NewCourseUnitObj extends UnitObject {
...
+    gradeInstances?: StudentGradeInstance[];
     overallBestScore: number = 0;
     effectiveScore: number = 0;
     bestScore: number = 0;
     numAttempts: number = 0;
     numLegalAttempts: number = 0;
     locked: boolean = false;
     currentProblemState?: unknown;
-    id?: number;
+    id: number = 0;
```
red-tt-bear, line 214: No, we should not be setting default values
Yes the nil checks are annoying, but all that is happening here is that we are disabling the typescript checks that protect us

We had a bug where course wasn't loading correctly because the id was defaulting to 0 so even though the course didn't load it tried to make the api call and fail

src/Courses/CourseInterfaces.tsx
```diff
@@ -87,14 +87,15 @@ enum TopicTypeId {
 export class TopicAssessmentFields {
     duration?: number;
     hardCutoff?: boolean;
-    maxGradedAttemptsPerRandomization?: number;
-    maxReRandomizations?: number;
-    randomizationDelay?: number;
+    maxGradedAttemptsPerVersion?: number;
+    maxVersions: number = 1;
```
red-tt-bear, line 91: I prefer no defaults (longer comment below)

src/Assignments/SimpleProblemPage.tsx
doombot-exe: it's an override for when a new version is started immediately following a submitted attempt -- 
* attempts remaining is `X`
* click submit, decrement attempts remaining (`setState(X-1)`), await scores
* results modal has 'end exam' option
* clicking 'end exam' must receive an "override" for attemptsRemaining because otherwise attempts remaining is still `X`

I'm not sure it's necessary here with versions in state -- but with the behavior seen for attemptsRemaining, I wanted to ensure accuracy.

src/Components/ModalWithChildren.tsx
doombot-exe: Modal with Children has been cancelled.

src/Components/ConfirmationModal.tsx
```diff
@@ -26,17 +29,38 @@ interface ComponentToggleButtonProps {
     | 'outline-info'
     | 'outline-dark'
     | 'outline-light';
+    secondaryVariant?: | 'primary'
```
PanopticaRising, line 32: Is that first `|` supposed to be there? Also, this is kinda long, maybe we should make a type for it?

src/Assignments/SimpleProblemPage.tsx
doombot-exe: 😝

src/Components/ConfirmationModal.tsx
doombot-exe: it's copypasta from the original `confirmVariant`

doombot-exe: updated

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -6,98 +6,305 @@ import ProblemIframe from './ProblemIframe';
...
-    useEffect(()=>{
+    useEffect(() => {
         setLoading(true);
         (async () => {
             try {
-                if(_.isNil(params.topicId)) {
+                if (_.isNil(params.topicId)) {
                     console.error('topicId is null');
                     setError('An unexpected error has occurred');
                     return;
```
PanopticaRising, line 52: If we return here, we never get out of the setLoading state, right? We should probably throw something instead.

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -6,98 +6,305 @@ import ProblemIframe from './ProblemIframe';
...
+                    if (!_.isNil(override)) {
+                        delete override.id;
+                        _.assign(prob, override);
+                    }
+                    return new ProblemObject(prob);
+                })
+                .keyBy('id')
+                .value();
+            setProblems(problemDictionary);
+            setSelectedProblemId(_.sortBy(problems, ['problemNumber'])[0].id);
```
PanopticaRising, line 92: https://lodash.com/docs/4.17.15#minBy
```suggestion
            setSelectedProblemId(_.minBy(problems, 'problemNumber').id);
```

src/Assignments/SimpleProblemPage.tsx
red-tt-bear: @doombot-exe is the best

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -6,98 +6,305 @@ import ProblemIframe from './ProblemIframe';
...
+            if (!_.isNil(topic.topicAssessmentInfo) &&
+                !_.isNil(topic.topicAssessmentInfo.studentTopicAssessmentInfo) &&
+                topic.topicAssessmentInfo.studentTopicAssessmentInfo.length > 0 // student has generated at least one version already
```
PanopticaRising, line 95: If Typescript infers the null correctly, this would be cleaner.
```suggestion
            if (!_.isNil(topic.topicAssessmentInfo?.studentTopicAssessmentInfo) &&
                !_.isEmpty(topic.topicAssessmentInfo.studentTopicAssessmentInfo) // student has generated at least one version already
```

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -6,98 +6,305 @@ import ProblemIframe from './ProblemIframe';
...
+            ) {
+                topic.topicAssessmentInfo.studentTopicAssessmentInfo = _.sortBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, ['startTime']).reverse();
```
PanopticaRising, line 97:
```suggestion
                topic.topicAssessmentInfo.studentTopicAssessmentInfo = _.sortBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, ['startTime'], ['desc']);
```

src/Assignments/SimpleProblemPage.tsx
doombot-exe: nifty!

src/Assignments/SimpleProblemPage.tsx
doombot-exe: this breaks other refs to topic.topicAssessmentInfo fields

src/Assignments/SimpleProblemPage.tsx
doombot-exe: this one I know *does* work ;)

src/Assignments/SimpleProblemPage.tsx
doombot-exe: TS falls apart because it claims ignorance of the fact that 'problemNumber' is a key for the ProblemObject[]

src/Assignments/SimpleProblemPage.tsx
doombot-exe: handled

src/Assignments/ProblemDetails.tsx
```diff
@@ -12,16 +12,23 @@ const INFINITE_MAX_ATTEMPT_VALUE = 0;
...
-    const startDate = moment(topic?.startDate);
-    const endDate = moment(topic?.endDate);
-    const deadDate = moment(topic?.deadDate);
-    const solutionsMoment = moment(deadDate).add(1, 'days');
+    const isAssessment = !_.isNil(topic?.topicAssessmentInfo?.studentTopicAssessmentInfo?.[0]);
+    const isClosed = isAssessment && topic?.topicAssessmentInfo?.studentTopicAssessmentInfo?.[0].isClosed;
+    const versionStartTime = topic?.topicAssessmentInfo?.studentTopicAssessmentInfo?.[0].startTime;
+    const versionEndTime = topic?.topicAssessmentInfo?.studentTopicAssessmentInfo?.[0].endTime;
+
+    const startDate = (_.isNil(versionStartTime)) ? moment(topic?.startDate) : moment(versionStartTime);
```
red-tt-bear, line 28: This could be simplified by doing:
```TypeScript
const startDate = versionStartTime?.toMoment() ?? topic?.startDate.toMoment()
```

Although I guess the toMoment extension is only in the backend it will be in the frontend with my PR

The problem with the way it is now is if topic is undefined startDate is inaccurately the current time

We can change that later

src/Assignments/ProblemDetails.tsx
```diff
@@ -75,14 +82,16 @@ export const ProblemDetails: React.FC<ProblemDetailsProps> = ({
...
                                     message = `Due ${endDate.fromNow()}`;
                                 } else if (currentMoment.isBefore(deadDate)) {
                                     message = `Partial credit expires ${deadDate.fromNow()}`;
                                 } else if (currentMoment.isBefore(solutionsMoment)) {
                                     message = `Solutions available ${solutionsMoment.fromNow()}`;
+                                } else if (isAssessment && (attemptsRemaining === 0 || isClosed)) {
+                                    message = 'is completed';
                                 } else {
-                                    message = 'Past due';
+                                    message = (isAssessment) ? 'Time expired' : 'Past due';
```
red-tt-bear, line 94: Does it make sense to just always make it `Time expired`
I don't think it really matters either way

src/Assignments/SimpleProblemPage.tsx
red-tt-bear: if we just
```TypeScript
throw new Error('An unexpected error has occurred');
```
then we don't even need to log it since it happens in the catch block

src/Assignments/SimpleProblemPage.tsx
red-tt-bear: 🤔

src/Assignments/ProblemDetails.tsx
doombot-exe: Well, the exam itself isn't necessarily "past due" -- perhaps this could be more elaborate? 
Perhaps "Time expired _for this version_"?

src/Assignments/SimpleProblemPage.tsx
doombot-exe: updated -- though in the catch block we end up performing the same action 😂

src/Assignments/SimpleProblemPage.tsx
red-tt-bear: Yeah, if I'm understanding the problem then this is one of the things that annoys me about typescript
If you use nil chaining and a nil check it doesn't determine all types in the chain to be not nil
For example
```Typescript
        const  r: Q | undefined = {};
        if (
            !_.isNil(r) &&
            !_.isNil(r.a) &&
            !_.isNil(r.a.b) &&
            !_.isNil(r.a.b.c)
        ) {
            console.log(r.a.b.c.d);
        }
        if (!_.isNil(r?.a?.b?.c)) {
            console.log(r.a.b.c.d);
        }
```

The above two cases should be equivalent but they are not, the bottom one says that all paths can be undefined except c (it doesn't check d since it isn't being chained)

red-tt-bear: If only I read @PanopticaRising's comment more closely 🤣 
> If Typescript infers the null correctly

src/Assignments/ProblemDetails.tsx
red-tt-bear: Hmm I think it's fine for now but I do like that

red-tt-bear (approved):

src/Assignments/ProblemDetails.tsx
```diff
@@ -12,16 +12,23 @@ const INFINITE_MAX_ATTEMPT_VALUE = 0;
...
 export const ProblemDetails: React.FC<ProblemDetailsProps> = ({
     problem,
     topic,
+    attemptsRemaining,
 }) => {
-    const startDate = moment(topic?.startDate);
-    const endDate = moment(topic?.endDate);
-    const deadDate = moment(topic?.deadDate);
-    const solutionsMoment = moment(deadDate).add(1, 'days');
+    const isAssessment = !_.isNil(topic?.topicAssessmentInfo?.studentTopicAssessmentInfo?.[0]);
```
PanopticaRising, line 23: This isn't strictly true, right? The true designator is what the topicTypeId is.

PanopticaRising: Unless we handle that filtering on the backend (but it wouldn't hurt to be defensive against it on the frontend)

src/Assignments/SimpleProblemPage.tsx
PanopticaRising: I prefer that we're consistent as well, eventually. We'll do a design review and try to land on something consistent.

PanopticaRising (approved):

src/Assignments/ProblemDetails.tsx
doombot-exe: you're right about this -- we should check topicTypeId *and* this is not nil, and change the name to `isVersionedAssessment` to be clear that we're checking for a version of the topic, which IS an assessment...

PR #199, opened 2020-10-21
Feature/red 144 exam submit

doombot-exe:
fixed bug where current version was not being used
fixed bug where time-expired exams allowed for answer submissions

### DIFF
--- a/src/Assignments/ProblemDetails.tsx
+++ b/src/Assignments/ProblemDetails.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { ProblemObject, TopicObject, StudentGrade } from '../Courses/CourseInterfaces';
+import { ProblemObject, TopicObject, StudentGrade, StudentTopicAssessmentFields } from '../Courses/CourseInterfaces';
 import _ from 'lodash';
 import moment from 'moment';
 import { OverlayTrigger, Tooltip, Badge } from 'react-bootstrap';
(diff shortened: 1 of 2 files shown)


src/Assignments/ProblemDetails.tsx
```diff
@@ -20,17 +20,40 @@ export const ProblemDetails: React.FC<ProblemDetailsProps> = ({
     topic,
     attemptsRemaining,
 }) => {
-    const isAssessment = !_.isNil(topic?.topicAssessmentInfo?.studentTopicAssessmentInfo?.[0]);
-    const isClosed = isAssessment && topic?.topicAssessmentInfo?.studentTopicAssessmentInfo?.[0].isClosed;
-    const versionStartTime = topic?.topicAssessmentInfo?.studentTopicAssessmentInfo?.[0].startTime;
-    const versionEndTime = topic?.topicAssessmentInfo?.studentTopicAssessmentInfo?.[0].endTime;
+    if (_.isNil(topic)) {
+        console.error('Problem details requested without a topic');
```
red-tt-bear, line 24: Should we be returning here?

src/Assignments/ProblemDetails.tsx
```diff
@@ -20,17 +20,40 @@ export const ProblemDetails: React.FC<ProblemDetailsProps> = ({
...
+        !_.isNil(topic.topicAssessmentInfo) &&
+        !_.isEmpty(topic?.topicAssessmentInfo?.studentTopicAssessmentInfo)
+    ) {
+        version = _.maxBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, 'startTime');
+        if (_.isNil(version)) {
+            console.error('We have versions, but an attempt to set the current version failed.');
+        }
+    }
+
+    const isVersionedAssessment = (topic?.topicTypeId === 2 && !_.isNil(version));
```
red-tt-bear, line 40: as @PanopticaRising pointed out we have an enum for this on the frontend, but I can fix this when I fix mine

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -91,12 +91,21 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
                 !_.isNil(topic.topicAssessmentInfo.studentTopicAssessmentInfo) &&
                 topic.topicAssessmentInfo.studentTopicAssessmentInfo.length > 0 // student has generated at least one version already
             ) {
-                topic.topicAssessmentInfo.studentTopicAssessmentInfo = _.sortBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, ['startTime'], ['desc']);
-                setAttemptsRemaining(topic.topicAssessmentInfo.studentTopicAssessmentInfo[0].maxAttempts! - topic.topicAssessmentInfo.studentTopicAssessmentInfo[0].numAttempts!);
+                const currentVersion = _.maxBy(topic.topicAssessmentInfo.studentTopicAssessmentInfo, 'startTime');
```
red-tt-bear, line 94: Shouldn't we sort this on the backend, doesn't hurt to have it in both places though

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -91,12 +91,21 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
+                if (!_.isNil(currentVersion) && !_.isNil(currentVersion?.numAttempts) && !_.isNil(currentVersion.maxAttempts)) {
+                    // if the assessment has expired - do NOT allow submissions...
```
red-tt-bear, line 96: We are blocking this on the backend as well right?

red-tt-bear (approved):

PanopticaRising (approved):

src/Assignments/ProblemDetails.tsx
doombot-exe: we cannot - react expects this component to always return from the same place?

src/Assignments/ProblemDetails.tsx
doombot-exe: I couldn't get the frontend ENUM to work for me - happy to receive a tutorial

src/Assignments/SimpleProblemPage.tsx
doombot-exe: This is a subinclude, which cannot be sorted

src/Assignments/SimpleProblemPage.tsx
doombot-exe: yes, but I'd rather not put the students in an error state

src/Assignments/ProblemDetails.tsx
red-tt-bear: You can return from other places, but react expects a jsx return
We could return something like:
```
return (<p>An error occurrend</p>)
```

You should also be able to return null or `(<></>)`

But it could also if we have an error handler we could throw an error, if it's further down you could move the try up

src/Assignments/SimpleProblemPage.tsx
red-tt-bear: subincludes can be sorted https://github.com/rederly/backend/blob/dev/src/features/courses/course-controller.ts#L80

src/Assignments/SimpleProblemPage.tsx
red-tt-bear: 💯

PR #203, opened 2020-10-22
Feature/red 180 client side logging

red-tt-bear:


doombot-exe (approved):
🌳 🪓

src/Utilities/BrowserConsoleLoggerTransport.ts
```diff
@@ -0,0 +1,43 @@
+import Transport from 'winston-transport';
+import _ from 'lodash';
+
+export default class BrowserConsoleLoggerTransport extends Transport {
```
PanopticaRising, line 4: There are tests in Winston that validate API compliance, might be useful in the future for debugging.

https://github.com/winstonjs/abstract-winston-transport/tree/master

src/Utilities/BrowserConsoleLoggerTransport.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    log(info: any, callback: any): void {
+        /**
+         * In the example they use setImmediate however that does not seem to have great browser support
+         * setTimouet 0 is equivalent
+         * Emitting the logged even seems to make login freeze up... we should look into why this is happening
+         * I don't know if there is any benefit to having it emit a logged event, unless we were observing it
+         */
+        setTimeout(() => {
+            this.emit('logged', info);
```
PanopticaRising, line 23: Looks like this is just to support things that are listening for logged events, potentialyl for other console extensions rather than any Winston-specific things.

https://github.com/winstonjs/winston/commit/f0d402426d03d587e57de3d18a0785030eb8a4db

PanopticaRising (approved):

PR #205, opened 2020-10-22
Feature/red 180 client side logging

red-tt-bear:


doombot-exe (approved):

src/Utilities/AxiosBatchTransport.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+interface LogMessage {
+    message: string;
+    time: Date;
+    level: string;
+    meta?: unknown;
+}
+
+export default class AxiosBatchTransport extends Transport {
```
PanopticaRising, line 19: My only concern with this is performance. Can the batching process run in the background? Does switching tabs affect the batching?

PanopticaRising (approved):

src/Utilities/AxiosBatchTransport.ts
red-tt-bear: I don't understand what you mean? It should be scoped to a tab and the batching process is just putting it into an array and sending them all at once

Technically I guess if you had two tabs open you could send twice as many requests but still I don't think that is a big deal

PR #207, opened 2020-10-22
update assessment status when time expires

doombot-exe:


src/Assignments/ProblemDetails.tsx
```diff
@@ -109,17 +111,19 @@ export const ProblemDetails: React.FC<ProblemDetailsProps> = ({
...
                                 } else if (currentMoment.isBefore(deadDate)) {
                                     message = `Partial credit expires ${deadDate.fromNow()}`;
                                 } else if (currentMoment.isBefore(solutionsMoment)) {
                                     message = `Solutions available ${solutionsMoment.fromNow()}`;
-                                } else if (isVersionedAssessment && (attemptsRemaining === 0 || isClosed)) {
-                                    message = 'is completed';
-                                } else {
+                                } else if (currentMoment.isAfter(solutionsMoment)) {
                                     message = (isVersionedAssessment) ? 'Time expired for this version' : 'Past due';
+                                    setAttemptsRemaining && setAttemptsRemaining(0);
```
PanopticaRising, line 122: This could be more simply written as 
```suggestion
                                    setAttemptsRemaining?.(0);
```

src/Assignments/ProblemDetails.tsx
```diff
@@ -13,12 +13,14 @@ interface ProblemDetailsProps {
     problem: ProblemObject;
     topic: TopicObject | null;
     attemptsRemaining?: number;
+    setAttemptsRemaining?: (a: number)=>unknown;
```
PanopticaRising, line 16: If this is a setState from useState, the type can be a bit more explicit, but that's not dreadfully important.

PanopticaRising (approved):

src/Assignments/ProblemDetails.tsx
red-tt-bear, line 122: Is this you using react lifecycle to use effect rather than just calling the method directly

Does this disrupt homework at all? I'd assume it calls a rerender but then nothing happens so it is fun

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -119,6 +119,7 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
             }
             
             if (actualVersionsRemaining > 0) {
+                // TODO do we need to check for start times?
```
red-tt-bear, line 122: What do you mean? right now the backend throws an error if you try to access it early so you would never get here

red-tt-bear (approved):

src/Assignments/SimpleProblemPage.tsx
doombot-exe: when the start assessment bug was found - I couldn't tell where I'd gone wrong. This was one possibility.

src/Assignments/ProblemDetails.tsx
doombot-exe: I could do `isVersionedAssessment && setAttemptsRemaining?.(0)`

red-tt-bear: Yeah, I don't think it is a big deal

PR #214, opened 2020-10-23
Bugfix/red 200 problem paths not showing up

PanopticaRising:


src/Courses/TopicSettings/ProblemSettings.tsx
```diff
@@ -117,7 +118,10 @@ export const ProblemSettings: React.FC<ProblemSettingsProps> = ({selected, setSe
             const newTopic = new TopicObject(topic);
             const newQuestion = _.find(newTopic.questions, ['id', selected.id]);
 
+            // TODO: Right now, the backend does not return the courseQuestionAssessmentInfo, so we should use the local data
+            // if the attempt was a success.
```
doombot-exe, line 122: I thought we found that courseQuestionAssessmentInfo _was_ being returned by the backend, no?

doombot-exe (approved):

src/Courses/TopicSettings/ProblemSettings.tsx
PanopticaRising: It was being returned from the GET call, but it's not returned from the PUT call (because of some finnickiness with the upsert operation)

red-tt-bear (approved):

PR #215, opened 2020-10-23
improved exam state handling and messages

doombot-exe:


src/Assignments/SimpleProblemPage.tsx
```diff
@@ -89,39 +89,67 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
+                    // determine status from current version, don't rely on useState variables to be immediately accurate
+                    const currentAttemptsRemaining = currentVersion.maxAttempts - currentVersion.numAttempts;
+                    let currentVersionsRemaining = versionsRemaining;
+                    if (!_.isNil(currentTopic.topicAssessmentInfo.maxVersions)) {
+                        currentVersionsRemaining = currentTopic.topicAssessmentInfo.maxVersions - currentTopic.topicAssessmentInfo.studentTopicAssessmentInfo.length;
+                        setVersionsRemaining(currentVersionsRemaining);
+                    }
+
+                    // if the assessment has been "closed" OR time has expired - do NOT allow submissions...
+                    if ((!_.isNil(currentVersion.isClosed) && currentVersion.isClosed) ||
```
red-tt-bear, line 107: isn't this first part just:
```
if (currentVersion.isClosed)
```

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -89,39 +89,67 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
+                        (!_.isNil(currentVersion.endTime) && currentVersion.endTime.toMoment().isBefore(moment()))
```
red-tt-bear, line 108: Could be simplified:
```
currentVersion.endTime?.toMoment().isBefore(moment())
```

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -89,39 +89,67 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
+                        }
                     } else {
-                        setAttemptsRemaining(0);
+                        setAttemptsRemaining(currentAttemptsRemaining);
                     }
-                }
-                if (!_.isNil(topic.topicAssessmentInfo.maxVersions)) {
-                    setVersionsRemaining(topic.topicAssessmentInfo.maxVersions - topic.topicAssessmentInfo.studentTopicAssessmentInfo.length);
+                } else {
+                    console.error('Something is wrong with this assessment version. There are versions provided, but no version is "current"');
```
red-tt-bear, line 134: Needs to be logger.error

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -89,39 +89,67 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
                 }
                 if (!_.isNil(currentVersion) && !_.isNil(currentVersion.id)) {
                     setVersionId(currentVersion.id);
                 }
-                setTopic(topic);
+                // setTopic(topic);
```
red-tt-bear, line 139: Why are we commenting this out?

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -379,12 +418,15 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
                                                 New Version
                                             </Button>
                                         }
-                                        <Button variant='danger'
-                                            tabIndex={0}
-                                            onClick={() => confirmEndVersion()}
-                                        >
-                                            End Exam
-                                        </Button>
+                                        {(attemptsRemaining!==topic.topicAssessmentInfo?.maxGradedAttemptsPerVersion) &&
```
red-tt-bear, line 421: Shouldn't use strict equals, should be
```
attemptsUsed < maxAttempts
```
in case there is a bug and they somehow get over the max attempts, then we want some damage control

red-tt-bear (changes requested):

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -379,12 +418,15 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
+                                            <Button variant='danger'
+                                                tabIndex={0}
+                                                onClick={() => confirmEndVersion()}
+                                                disabled={attemptsRemaining===topic.topicAssessmentInfo?.maxGradedAttemptsPerVersion}
```
red-tt-bear, line 425: Same thing goes for this disabled

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -379,12 +418,15 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
-                                        <Button variant='danger'
-                                            tabIndex={0}
-                                            onClick={() => confirmEndVersion()}
-                                        >
-                                            End Exam
-                                        </Button>
+                                        {(attemptsRemaining!==topic.topicAssessmentInfo?.maxGradedAttemptsPerVersion) &&
+                                            <Button variant='danger'
+                                                tabIndex={0}
+                                                onClick={() => confirmEndVersion()}
```
red-tt-bear, line 424: No need to pass a function that calls a function
```
onClick={confirmEndVersion}
```

src/Assignments/SimpleProblemPage.tsx
doombot-exe: setTopic is already handled above -- this was an artifact that didn't get cleaned up

src/Assignments/SimpleProblemPage.tsx
doombot-exe: oops, that disabled was supposed to GO... replaced by the logic that simply removes the button

src/Assignments/SimpleProblemPage.tsx
doombot-exe: I do, because confirmEndVersion has optional arguments.

red-tt-bear (approved):

PR #231, opened 2020-10-26
Bugfix/red 204 preview blocks submit

PanopticaRising:


src/Courses/TopicSettings/ProblemSettings.tsx
```diff
@@ -304,9 +185,11 @@ export const ProblemSettings: React.FC<ProblemSettingsProps> = ({selected, setSe
                             </Grid>
                         )}
                     </Grid><Grid item xs={12}>
-                        {
-                            rendererPreview()
-                        }
+                        <RendererPreview 
+                            defaultPath={topic.topicTypeId === TopicTypeId.EXAM ? 
+                                additionalProblemPaths?.[0].path || '' :
```
doombot-exe, line 190: don't we have something better to fall back to besides empty-string? Will there be a value for webworkQuestionPath if the topicType is EXAM?

doombot-exe (approved):

red-tt-bear (approved):

src/Courses/TopicSettings/ProblemSettings.tsx
red-tt-bear: If the array is empty this would result in a null pointer exceptions?

PR #236, opened 2020-10-26
display maxVersions with overrides

doombot-exe:


src/Assignments/SimpleProblemPage.tsx
```diff
@@ -70,8 +70,20 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
 
         const currentTopic = res.data.data.topic;
 
-        if (currentTopic.studentTopicOverride?.length > 0) {
-            _.assign(currentTopic, currentTopic.studentTopicOverride[0]);
+        // apply topic overrides: start/end/dead dates
+        if (!_.isNil(currentTopic.studentTopicOverride) && !_.isEmpty(currentTopic.studentTopicOverride)) {
+            const override = currentTopic.studentTopicOverride[0];
+            delete override.id;
```
red-tt-bear, line 76: This is not a good way to do it
In some cases there are foreign keys that are in common as well
We should keep this consistent and have an interface for the data portion of the overrides table and then do the assign based on a pick

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -70,8 +70,20 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
+            _.assign(currentTopic, override);
+        }
+        // apply assessment overrides: duration, maxAttempts, maxVersions, versionDelay
+        // only maxVersions is relevant -- the others are accounted for in version (studentTopicAssessmentInfo)
+        if (!_.isNil(currentTopic.topicAssessmentInfo) && 
+            !_.isNil(currentTopic.topicAssessmentInfo.studentTopicAssessmentOverride) && 
+            !_.isEmpty(currentTopic.topicAssessmentInfo.studentTopicAssessmentOverride)
```
red-tt-bear, line 83: isEmpty checks nil, I don't know if strict will accept it as a nil check as well though, in theory you might not need the isNil check

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -70,8 +70,20 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
+        ) {
+            const override = currentTopic.topicAssessmentInfo?.studentTopicAssessmentOverride[0];
+            currentTopic.topicAssessmentInfo.maxVersions =  override.maxVersions;
```
red-tt-bear, line 86: This will leave us in a false sense of security, on the other overrides if you add a field it should just work
If we stop doing it with the pick and assign additional fields added won't `just work`

red-tt-bear (changes requested):
I think it is important we stay consistent with how we do overrides, it'll make it easier to find where override logic is done and it has been set up to expand automatically elsewhere

src/Assignments/SimpleProblemPage.tsx
red-tt-bear: This is NIT btw

doombot-exe: it didn't pass without the additional check, believe me - I'd rather this were shorter

src/Assignments/SimpleProblemPage.tsx
red-tt-bear: Ahh I thought this was the backend for a second, I guess we never really did it right on the frontend

red-tt-bear (approved):

PR #242, opened 2020-10-28
Added an EmailProfessor component that renders an email modal and sen…

PanopticaRising:
…ds the email along.

### DIFF
--- a/src/Assignments/EmailProfessor.tsx
+++ b/src/Assignments/EmailProfessor.tsx
@@ -14,7 +14,7 @@ interface EmailProfessorProps {
 /**
  * This is a button-and-model pair.
  */
-export const EmailProfessor: React.FC<EmailProfessorProps> = ({topic, problem}) => {
+export const EmailProfessor: React.FC<EmailProfessorProps> = ({problem}) => {
     const {course} = useCourseContext();
     const [content, setContent] = useState<string>('');
     const [show, setShow] = useState<boolean>(false);
@@ -27,12 +27,6 @@ export const EmailProfessor: React.FC<EmailProfessorProps> = ({topic, problem}) 
                 content: content,
                 question: {
                     id: problem.id,
-                    problemNumber: problem.problemNumber,
-                    webworkQuestionPath: problem.webworkQuestionPath,
-                },
-                topic: {
-                    id: topic.id,
-                    name: topic.name,
                 }
             });
             setSendEmailRespMsg({message: res.data.message ?? 'Email sent.', variant: 'success'});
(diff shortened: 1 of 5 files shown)


src/Assignments/EmailProfessor.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+    const onClick = async () => {
+        try {
+            const res = await postEmailProfessor({
+                courseId: course.id,
+                content: content,
+                question: {
+                    id: problem.id,
+                    problemNumber: problem.problemNumber,
+                    webworkQuestionPath: problem.webworkQuestionPath,
```
red-tt-bear, line 31: remove this please

red-tt-bear:
Looks good, just need to remove the parameters that are listed in the backend PR

src/Assignments/EmailProfessor.tsx
```diff
@@ -0,0 +1,73 @@
...
+            const res = await postEmailProfessor({
+                courseId: course.id,
+                content: content,
+                question: {
+                    id: problem.id,
+                }
+            });
+            setSendEmailRespMsg({message: res.data.message ?? 'Email sent.', variant: 'success'});
+        } catch (e) {
+            console.error(e);
```
red-tt-bear, line 34: 👿 👿 👿 👿 👿

red-tt-bear: This isn't a nit anymore since we have the logger implemented, especially since it is an error this won't bubble up
If we think it can validly fail then we should change it to a warn

src/Assignments/EmailProfessor.tsx
```diff
@@ -0,0 +1,73 @@
...
+                                onChange={(e: any) => setContent(e.target.value)}
+                            />
+                        </FormGroup>
+                    </Form>
+                </ModalBody>
+                <ModalFooter>
+                    <Button 
+                        variant="primary" 
+                        onClick={onClick}
+                        // disabled={sendEmailRespAlertType === 'success'}
```
red-tt-bear, line 64: Are we disabling the button while the request is going out and while there are no changes?

red-tt-bear (changes requested):
console.error and make sure we can't spam professors by accident

src/Assignments/EmailProfessor.tsx
PanopticaRising: We should, I meant to go back and add this in when I had validation setup.

For preventing spamming, we have a bigger issue about throttling the calls on the backend, since a student could easily force a bunch of requests.

src/Assignments/EmailProfessor.tsx
```diff
@@ -0,0 +1,83 @@
...
+                        <FormGroup>
+                            <FormLabel>Compose your email below:</FormLabel>
+                            <FormControl 
+                                as='textarea' 
+                                size='sm' style={{height: '200px'}}
+                                autoComplete='off'
+                                onChange={(e: any) => {
+                                    setContent(e.target.value);
+                                    if (disabled) {
+                                        setDisabled(false);
```
doombot-exe, line 62: do we want to confirm that the contents are not empty before disable -> false?

doombot-exe (approved):

PR #248, opened 2020-10-31
Feature/red 232 integrated editor

red-tt-bear:


src/Assignments/ProblemEditor.tsx
```diff
@@ -0,0 +1,391 @@
...
+            message: '',
+            variant: 'info'
+        });
+    };
+
+    useEffect(() => {
+        (async () => {
+            const loadPromise = load();
+            const catalogPromise = getCatalog();
+            await Promise.all([loadPromise, catalogPromise]);
```
PanopticaRising, line 163: Why are you awaiting a promises here? Do they not complete if you don't wait for them?

PanopticaRising (approved):

src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
```diff
@@ -343,7 +347,21 @@ export const postPreviewQuestion = async ({
...
+    showHints,
+    showSolutions
 }: PreviewQuestionOptions): Promise<AxiosResponse<GetQuestionResponse>> => {
+    if (_.isNil(problemSource) === _.isNil(webworkQuestionPath)) {
+        throw new Error('Either problem source or webwork question path must be defined, not neither, not both');
+    }
+
+    if (!_.isNil(problemSource)) {
+        formData = formData ?? new FormData();
+        formData.set('problemSource', Buffer.from(problemSource).toString('base64'));
```
doombot-exe, line 360: just something to tag for the future: rederly/renderer#7
we will need to re-think this encoding for int'l support

src/Assignments/ProblemEditor.tsx
red-tt-bear: They do, I added this for supporting the loading spinner but never got to the loading spinner and it didn't seem to need it, in the future we should probably throttle the website and see what it looks like when requests aren't so quick

PR #255, opened 2020-11-02
Bugfix/red 259 topic grading state mgmt

doombot-exe:


src/Courses/TopicGrades/GradingPage.tsx
```diff
@@ -31,7 +31,7 @@ export const TopicGradingPage: React.FC<TopicGradingPageProps> = () => {
     const [gradeOverride, setGradeOverride] = useState<Partial<StudentGrade>>({});
     const [isPinned, setIsPinned] = useState<Pin | null>(null); // pin one or the other, not both
     const [topic, setTopic] = useState<TopicObject | null>(null);
-    const [problems, setProblems] = useState<ProblemObject[] | null>(null);
+    const [problems, setProblems] = useState<NewProblemObject[] | null>(null);
```
PanopticaRising, line 34: We're still using `NewProblemObject`? Does it have something the regular `ProblemObject` doesn't have?

PanopticaRising (approved):

src/Courses/TopicGrades/GradingPage.tsx
doombot-exe: extends ProblemObject and includes courseTopicContentId -- which I ended up using as a test to determine whether or not we need to re-fetch our problems and/or re-map those problems...

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -210,18 +210,18 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
 
     const confirmEndVersion = (actualAttemptsRemaining?: number) => {
         actualAttemptsRemaining = actualAttemptsRemaining ?? attemptsRemaining;
-        let message = '';
+        let message = 'You have successfully completed this exam.';
```
red-tt-bear, line 213: I personally prefer not to have the success message as the fallback
If there was an unaccounted error it would say success 🤣

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -76,8 +76,10 @@ export const GradeInfoHeader: React.FC<GradeInfoHeaderProps> = ({
         let currentAttemptsCount: number | undefined;
         let currentAverageScore: number | undefined;
         let currentVersionMap: Record<number, Array<number>> = {};
+        let versioned = false;
         if (!_.isNil(workbooks) && !_.isEmpty(workbooks)) {
             const workbookKeys = Object.keys(workbooks);
+            versioned = !_.isNil(Object.values(workbooks)[0].studentGradeInstanceId);
```
red-tt-bear, line 82: Do we have to worry about workbooks being an empty object?

red-tt-bear (approved):

src/Courses/TopicGrades/GradeInfoHeader.tsx
doombot-exe: two lines up, workbooks "not-isEmpty check"

src/Assignments/SimpleProblemPage.tsx
doombot-exe: next line, `if (actualAttemptsRemaining > 0)` ;P

red-tt-bear: 🤣 Sounds good should be fine
I would just do it the opposite way in case things change in the future, I want success to mean everything succeeded rather than nothing failed, for now, this is fine though

red-tt-bear (approved):

PR #256, opened 2020-11-02
RED-260 Added query parameter for load path

red-tt-bear:


doombot-exe (approved):

src/Extensions/StringExtension.ts
```diff
@@ -0,0 +1,22 @@
...
+    }    
+}
+
+/**
+ * AS STRING
+ * String prototype has this of type `String`, the capitalization causes issues so I have to cast as `string`
+ */
+String.prototype.toBase64 = function (): string {
+    // see AS STRING comment
+    return btoa(this as string);
```
PanopticaRising, line 14: Do you want to add the additional uriencode call to ensure that the `=` don't cause problems?

PanopticaRising (approved):

src/Extensions/StringExtension.ts
red-tt-bear: I was going to, however URLSearchParams from useQuery that we added and window.open handle that for us

PR #259, opened 2020-11-03
Feature/export all

PanopticaRising:


src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
```diff
@@ -585,3 +586,13 @@ export const deleteAttachments = async ({
         throw new BackendAPIError(e);
     }
 };
+
+export const getAllContentForVersion = async ({
+    gradeInstanceId
+}: {gradeInstanceId: number}): Promise<AxiosResponse<any>> => {
```
red-tt-bear, line 592: This is bad, we should try not to make a habbit of it (using any and typing inline)
Also even if we are rushing it should be `BackendAPIResponser<any>`

src/Assignments/ProblemIframe.tsx
```diff
@@ -346,7 +346,7 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
                     }
                 }}
                 title='Problem Frame'
-                style={{width: '100%', height: height, border: 'none', minHeight: '350px', visibility: (loading || error) ? 'hidden' : 'visible'}}
+                style={{width: '100%', height: height, border: 'none', minHeight: readonly ? '' : '350px', visibility: (loading || error) ? 'hidden' : 'visible'}}
```
red-tt-bear, line 349: Shouldn't it be `undefiend : 350px`
I've definitely weird issues with this before, I think if that didn't work you could _.omitBy(styles, _.isUndefined)
The problem with this (which is not an issue now) is I would think you would overwrite it with an invalid style, so if something else styled it you would lose that styling when you just didn't want it applied

src/Courses/TopicGrades/AttachmentsPreview.tsx
```diff
@@ -41,7 +43,14 @@ export const AttachmentsPreview: React.FC<AttachmentsPreviewProps> = ({gradeId,
 
     return (
         <Grid container style={{paddingLeft: '1rem'}}>
-            <Grid item md={12}><h1>Attachments</h1></Grid>
+            <Grid item md={12}><h1>Attachments</h1>
+                <PrintingPage 
+                    debug={true}
```
red-tt-bear, line 48: Suggestion:
process.env.NODE_ENV !== 'production'

src/Courses/TopicGrades/PrintEverything.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+    return (
+        <>
+            <h1>Grades for {gradeData[0].user.firstName} {gradeData[0].user.lastName}</h1>
+            <h2 className='dont-print'>Printing will begin in several seconds...</h2>
+            {gradeData.map((problem: any)=>{
+                const bestAttempt = problem.bestVersionAttempt;
+                const attachments = problem.studentGradeInstanceProblemAttachments;
+                // TODO: Get from call
+                const baseUrl = 'https://staging.rederly.com/work/';
```
red-tt-bear, line 48: what is this...

red-tt-bear (approved):

src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
```diff
@@ -589,10 +589,33 @@ export const deleteAttachments = async ({
 };
 
 export const getAllContentForVersion = async ({
-    gradeInstanceId
-}: {gradeInstanceId: number}): Promise<AxiosResponse<any>> => {
+    gradeId
+}: {gradeId: number}): Promise<AxiosResponse<any>> => {
     try {
-        return await AxiosRequest.get(url.resolve(COURSE_VERSION_PATH, `${gradeInstanceId}/`));
+        console.log(gradeId);
```
red-tt-bear, line 595: 👿

red-tt-bear (approved):

PR #267, opened 2020-11-03
Bugfix/fix swapped labels

PanopticaRising:


red-tt-bear (approved):

src/Courses/TopicSettings/GenericFormInputs.tsx
```diff
@@ -221,7 +221,7 @@ export const MaxVersionsField: React.FC<{}> = () => {
             name={`${examFieldNamePrefix}.maxVersions`}
             InputLabelProps={{ shrink: true }}
             inputRef={register()}
-            label={'Number of Versions'}
+            label={'Max Versions'}
```
doombot-exe, line 224: I don't understand why this is changed from 'Number of Versions' to 'Max Versions' -- "max" is definitely shorthand

doombot-exe (approved):

PR #268, opened 2020-11-04
Feature/red 242 cache busting

red-tt-bear:


src/Utilities/VersionCheck.tsx
```diff
@@ -0,0 +1,70 @@
...
+            let shouldReload = false;
+            if (_.isNil(versionCheckDate)) {
+                shouldReload = true;
+                logger.debug('Version Check: Time to try');
+            } else if (moment().isAfter(moment(versionCheckDate))) {
+                shouldReload = true;
+                logger.debug(`Version Check: Time to try again ${versionCheckDate}`);
+            } else {
+                shouldReload = false;
+                logger.error(`Version Check: Version mismatch packageVersion ${packageVersion} !== versionResponse ${versionResponse}. Too soon to try again ${versionCheckDate}`);
```
PanopticaRising, line 36: This error seems like it would bloat the logs quickly.

red-tt-bear: Nope, this can happen at most once per day per user (unless they clear all of their local storage but are still loading the cached page)
And in theory this can only happen after around realeases
This does not happen on the initial failure either, it is specifically after a hard reload failed

PanopticaRising: So this is an error that can happen? Not TSNH? 🤔 

I don't think it's an error if a user clears their cache, that seems like a warning at most, but I'll approve it if you disagree

PanopticaRising (approved):

src/Utilities/VersionCheck.tsx
red-tt-bear: Yeah I agree I think it should probably be a warning

That being said I would like this to be sent up to the server at least for now to gauge the solution (based on the jira comment)

Maybe it also makes sense to push warnings up to the server 🤷‍♂️ 

I would leave it for now until we figure out the above comment

doombot-exe (approved):

PR #278, opened 2020-11-10
add loading state to statistics tab

doombot-exe:
statistics tab renders the table before statistics data is updated - causing mismatch in rendered rows
**problem**: when drilling down to the attempts level, users may attempt to open an attempt before data is ready
**solution**: add a loading spinner for the duration of the data retrieval

### DIFF
--- a/src/Courses/CourseDetailsTabs/StatisticsTab.tsx
+++ b/src/Courses/CourseDetailsTabs/StatisticsTab.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/display-name */
 import React, { useState, forwardRef, useEffect } from 'react';
-import { Alert, Button, Col, Nav, Spinner } from 'react-bootstrap';
+import { Alert, Button, Col, Nav } from 'react-bootstrap';
 import MaterialTable, { Column } from 'material-table';
 // import { MdSearch, MdFirstPage, MdLastPage, MdClear, MdFilterList, MdChevronRight, MdChevronLeft, MdArrowDownward, MdFileDownload} from 'react-icons/md';
 import { Clear, SaveAlt, FilterList, FirstPage, LastPage, ChevronRight, ChevronLeft, Search, ArrowDownward } from '@material-ui/icons';
@@ -18,6 +18,7 @@ import { IAlertModalState } from '../../Hooks/useAlertState';
 import { putQuestionGrade } from '../../APIInterfaces/BackendAPI/Requests/CourseRequests';
 import { EnumDictionary } from '../../Utilities/TypescriptUtils';
 import logger from '../../Utilities/Logger';
+import { CircularProgress } from '@material-ui/core';
 
 const FILTERED_STRING = '_FILTERED';
 
@@ -517,7 +518,7 @@ export const StatisticsTab: React.FC<StatisticsTabProps> = ({ course, userId }) 
                     />
                 </>}
                 { loading ? 
-                    <Spinner animation='border' role='status'><span className='sr-only'>Loading...</span></Spinner> :
+                    <CircularProgress /> :
                     <MaterialTable
                         icons={icons}
                         title={(
--- a/src/Courses/CourseDetailsTabs/StatisticsTab.tsx
+++ b/src/Courses/CourseDetailsTabs/StatisticsTab.tsx
@@ -1,6 +1,6 @@
 /* eslint-disable react/display-name */
 import React, { useState, forwardRef, useEffect } from 'react';
-import { Alert, Button, Col, Nav } from 'react-bootstrap';
+import { Alert, Button, Col, Nav, Spinner } from 'react-bootstrap';
 import MaterialTable, { Column } from 'material-table';
 // import { MdSearch, MdFirstPage, MdLastPage, MdClear, MdFilterList, MdChevronRight, MdChevronLeft, MdArrowDownward, MdFileDownload} from 'react-icons/md';
 import { Clear, SaveAlt, FilterList, FirstPage, LastPage, ChevronRight, ChevronLeft, Search, ArrowDownward } from '@material-ui/icons';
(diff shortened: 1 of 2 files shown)


src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -137,6 +138,7 @@ export const StatisticsTab: React.FC<StatisticsTabProps> = ({ course, userId })
     const globalView = statisticsViewFromAllStatisticsViewFilter(view);
 
     useEffect(() => {
+        logger.silly(`Stats tab: [useEffect] course.id ${course.id}, globalView ${globalView}, idFilter ${idFilter}, userId ${userId}, userType ${userType}`);
```
red-tt-bear, line 141: We need some cut and dry reasoning for using silly
Right now the only issue is that console does not have a silly log so it will log with console.log

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -222,15 +225,19 @@ export const StatisticsTab: React.FC<StatisticsTabProps> = ({ course, userId })
                         completionPercent: formatNumberString(d.completionPercent, true)
                     }));
                 }
+                logger.silly('Stats tab: [useEffect] setting rowData');
                 setRowData(data);
+                setLoading(false);
             } catch (e) {
                 logger.error('Failed to get statistics.', e);
+                setLoading(false);
```
red-tt-bear, line 233: We could remove the below return and put the setLoading outside the try catch

Better yet we could put setLoading in a finally block

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -497,63 +517,68 @@ export const StatisticsTab: React.FC<StatisticsTabProps> = ({ course, userId })
...
-
-                                <Button
-                                    variant={grade.locked ? 'warning' : 'danger'}
-                                    className="ml-1 mr-1"
-                                    onClick={() => setGradesState({
-                                        ...gradesState,
-                                        view: GradesStateView.LOCK
-                                    })}
+                { loading ? 
+                    <Spinner animation='border' role='status'><span className='sr-only'>Loading...</span></Spinner> :
```
red-tt-bear, line 521: Didn't @PanopticaRising add a global loading spinner?

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -497,63 +517,68 @@ export const StatisticsTab: React.FC<StatisticsTabProps> = ({ course, userId })
...
+
+                                    <Button
+                                        variant={grade.locked ? 'warning' : 'danger'}
+                                        className="ml-1 mr-1"
+                                        onClick={() => {
+                                            logger.silly('Stats tab: [table button] setting gradesstate');
+                                            setGradesState({
+                                                ...gradesState,
+                                                view: GradesStateView.LOCK
+                                            });}}
```
red-tt-bear, line 558: NIT: IMO we should have all opening closures have closing closures at the same indentation
You can do multiple opening and closing on the same line however if you have 2 openings togeter you should have 2 closing

In this example there is an opening closure for onClick prop, then another one for method start
However on this last line the setGradesState parenthesis and bracket are there as well

It makes it harder to read however if we get a formatter in here it will automically be cleaned up

red-tt-bear (approved):

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
doombot-exe: okay, so info then? This was just an attempt to follow the render cycle for this component

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
doombot-exe: done

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
doombot-exe: done

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
```diff
@@ -137,6 +138,7 @@ export const StatisticsTab: React.FC<StatisticsTabProps> = ({ course, userId })
     const globalView = statisticsViewFromAllStatisticsViewFilter(view);
 
     useEffect(() => {
+        logger.info(`Stats tab: [useEffect] course.id ${course.id}, globalView ${globalView}, idFilter ${idFilter}, userId ${userId}, userType ${userType}`);
```
PanopticaRising, line 141: We should consider making a PREFIX variable, just in case someone mistypes the prefix.

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
doombot-exe: swap to material-ui CircularProgress

PanopticaRising (approved):

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
PanopticaRising: @red-tt-bear I never added the functionality for it to actually function globally.

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
PanopticaRising: @red-tt-bear if we're going to balk at using custom log-levels because they all log as Log, then we probably should remove them and stick to the browser defaults.

I think the reasoning for silly is straightforward- it's a log that is verbose or unlikely to be helpful.

src/Courses/CourseDetailsTabs/StatisticsTab.tsx
red-tt-bear: Yeah I'm between two approaches here:
1. Add a tag variable like android logcat does
2. Create logging interfaces that add the tag for you
  * I wanted this to automatically happen however I ran into issues with the source map
  * I know typescript types are stripped but I wonder if you could get something like a toString and bind it to the logger

PR #279, opened 2020-11-10
Feature/red 281 pg editor upload

red-tt-bear:


src/Enums/UserRole.tsx
```diff
@@ -9,12 +9,14 @@ export enum UserRole {
...
+export const unauthorizedRedirect = (doRedirect: boolean = true) => {
     // TODO: Generic redirect to handle clearing cookies.
     general.loginRedirectURL = `${window.location.pathname}${window.location.search}`;
     Cookies.remove(CookieEnum.USERTYPE);
-    Cookies.remove(CookieEnum.SESSION);    
-    window.location.assign('/');
+    Cookies.remove(CookieEnum.SESSION);
+    if (doRedirect) {
+        window.location.assign('/');        
+    }
```
doombot-exe, line 19: 👍

doombot-exe (approved):

src/Enums/UserRole.tsx
PanopticaRising: When do we want to remove the session without redirection to login?

PanopticaRising (approved):

src/Enums/UserRole.tsx
red-tt-bear: @PanopticaRising react router returns a redirect component which does not cause another page reload
The code that was here did a page redirect
Since I made the unauthorized logic used in multiple places you would get both happening in some cases and the page would flicker

This is not the final solution, I would prefer doing something with global state that returned the redirect a higher level

PR #280, opened 2020-11-10
RED-254 Added users from the course to guard logs when a user that's not

PanopticaRising:
enrolled in the course views the topic.


doombot-exe (approved):

doombot-exe (approved):

red-tt-bear:
Looks good

PR #281, opened 2020-11-11
problem iframe formURL mismatch

doombot-exe:
@PanopticaRising was clutch on this fix

### DIFF
--- a/src/Assignments/ProblemIframe.tsx
+++ b/src/Assignments/ProblemIframe.tsx
@@ -23,6 +23,11 @@ interface ProblemIframeProps {
     readonly?: boolean;
 }
 
+interface PendingRequest {
+    cancelled?: boolean;
+    problemId?: number;
+}
+
 /**
  * The most important part- rendering the problem.
  * We used the document.write strategy before for backwards compatibility, but modern browsers now block it.
@@ -31,7 +36,7 @@ interface ProblemIframeProps {
  * Important reference: https://medium.com/the-thinkmill/how-to-safely-inject-html-in-react-using-an-iframe-adc775d458bc
  */
 export const ProblemIframe: React.FC<ProblemIframeProps> = ({
-    problem: problemProp,
+    problem,
     previewPath,
     previewSeed,
     previewProblemSource,
@@ -42,8 +47,8 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
     readonly = false,
 }) => {
     const iframeRef = useRef<IFrameComponent>(null);
+    const pendingReq = useRef<PendingRequest | null>(null);
     const [renderedHTML, setRenderedHTML] = useState<string>('');
-    const [problem, setProblem] = useState<ProblemObject>(problemProp);
     const [loading, setLoading] = useState(true);
     const [error, setError] = useState('');
     const [lastSubmission, setLastSubmission] = useState({});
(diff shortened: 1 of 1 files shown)


src/Assignments/ProblemIframe.tsx
```diff
@@ -51,7 +52,34 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+                if (!_.isNil(rendererHTML) && 
+                    (!_.isNil(previewPath) || 
+                    !_.isNil(previewProblemSource) || 
+                    !_.isNil(match))
+                ) {
+                    // we match or else we're in a situation where there's nothing relevant to match *to*
+                    setRenderedHTML(rendererHTML);
+                } else {
+                    // do we want any notification of our state conflict or should we ditch this?
+                    if (!_.isNil(rendererHTML)) logger.warn('Someone has been rifling through the problems too quickly!');
```
PanopticaRising, line 77: This is probably better as a debug since we turned on warning notifications. We will get messages about this.

PanopticaRising (approved):
This is good, just change the logger severity cause we'll get lots of notifs about that.

PanopticaRising (approved):

src/Assignments/ProblemIframe.tsx
```diff
@@ -60,42 +89,46 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
-                }
-                // TODO: Error handling.
-                setRenderedHTML(res.data.data.rendererData.renderedHTML);
-            } catch (e) {
-                setError(e.message);
-                logger.error('Error posting preview', e, e.message);
-                setLoading(false);
-            }
-        })();
+        fetchHTML();
```
red-tt-bear, line 93: Why did we move this to another function if it is never reused and also scoped so that it can't be reused?

red-tt-bear, line 93: Also this would have been a problem before (and/or is not a problem) but is there a reason we have an orphaned async function and then more logic after it?

src/Assignments/ProblemIframe.tsx
```diff
@@ -60,42 +89,46 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
         setLastSubmission({});
     }, [problem, problem.id, workbookId, previewPath, previewProblemSource, previewSeed]);
 
+    const getHTML = async () => {
+        try {
+            let queryString = qs.stringify(_({
+                workbookId,
+                readonly
+            }).omitBy(_.isUndefined).value());
+            if (!_.isEmpty(queryString)) {
```
red-tt-bear, line 107: Nit: I don't know if we areally need an empty check here, having an empty query string should be fine

src/Assignments/ProblemIframe.tsx
PanopticaRising: Just on that point, it's a good thing to break code out into separate functions, even if they're not reusable. If helps readability and improves cyclomatic complexity scores.

red-tt-bear (approved):

src/Assignments/ProblemIframe.tsx
red-tt-bear: That's a great point, I like it

Still feels like we should move it out of the really finite scope of useEffect, no reason it has to be in here

PR #282, opened 2020-11-11
Bugfix/red 246 error posting preview

PanopticaRising:


src/APIInterfaces/BackendAPI/BackendAPIError.ts
```diff
@@ -33,6 +34,7 @@ export default class BackendAPIError extends Error {
         // Super needs to be called before I can call this
         if (isAxiosError(e)) {
             (this.data as any) = e.response?.data?.data;
+            this.status = e.response?.status;
```
red-tt-bear, line 37: I'm back and forth here, I don't like copying the data

What if we made a getter call `getAxiosError` That returned AxiosError or undefined
the code would look like

```
get axiosError(): AxiosError | null {
if (isAxiosError(this.originalError)) {
return this.originalError;
}
return null;
}

usage would be:
error.axiosError?.status;
```

The main reason I like it is if people are accustomed to working with axios attributes there is other data they could grab

src/Assignments/ProblemIframe.tsx
```diff
@@ -118,13 +110,21 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+                    readonly
+                });
             }
             return res.data.data.rendererData.renderedHTML as string;
 
         } catch (e) {
             setError(e.message);
-            logger.error('Error posting preview', e, e.message);
+            
+            if (_.isNil((e as BackendAPIError).status) || ((e as BackendAPIError).status !== 200 && (e as BackendAPIError).status !== 400)) {
```
red-tt-bear, line 124: Why don't we do instance checking here so we don't have to keep casting?
You could add this method to BackendAPIError

```Typescript
    static isBackendAPIError(obj: unknown): obj is BackendAPIError {
        return obj instanceof BackendAPIError;
    }
```

or you could just do the instance check here

src/Courses/CourseInterfaces.tsx
```diff
@@ -323,6 +323,7 @@ export class ProblemObject implements IProblemObject {
         additionalProblemPaths?: Array<string>;
         randomSeedSet?: number[];
     }
+    rendererData?: any;
```
red-tt-bear, line 326: What is this doing here? It doesn't look like it's used

src/APIInterfaces/BackendAPI/BackendAPIError.ts
red-tt-bear: Just quickly threw it in without testing but this seems like it would be the code:
```Typescript
get axiosError(): AxiosError | null {
    if (isAxiosError(this.originalError)) {
        return this.originalError;
    }
    return null;
}
```

And this would be the usage:
```
const temp = new BackendAPIError(new Error('tomtom'));
console.log(temp.axiosError?.response?.status);
```

PanopticaRising: Sure, I'll add that in, but then, does it make sense to add a getter for Message and instead of having us support all types of Errors? Even if we swapped out Axios, I would expect a BackendAPIError to have a message and a status.

red-tt-bear: That's a good point

I don't know if we can though since it is variable on super

src/Courses/CourseInterfaces.tsx
PanopticaRising: It's being used in the problemIframe now that the returned data from GetQuestion is being properly typed.

doombot-exe (approved):

src/Assignments/ProblemIframe.tsx
```diff
@@ -118,13 +113,21 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
+                    readonly
+                });
             }
             return res.data.data.rendererData.renderedHTML as string;
 
         } catch (e) {
             setError(e.message);
-            logger.error('Error posting preview', e, e.message);
+            
+            if (!BackendAPIError.isBackendAPIError(e) || (e.status !== 200 && e.status !== 400)) {
```
red-tt-bear, line 127:
```
e.status !== 200 && e.status !== 400
```
Seems confusing to me but I might be overthinking it, I checked and it works as expected 🤷‍♂️

red-tt-bear (approved):

PR #283, opened 2020-11-11
elaborate on problemMainForm error

doombot-exe:
added renderedHTML to the error log, and styled the error message (+ instructions to refresh)

### DIFF
--- a/src/Assignments/ProblemIframe.tsx
+++ b/src/Assignments/ProblemIframe.tsx
@@ -1,7 +1,7 @@
 import React, { useRef, useEffect, useState } from 'react';
 import { ProblemObject } from '../Courses/CourseInterfaces';
 import _ from 'lodash';
-import { Spinner } from 'react-bootstrap';
+import { Alert, Spinner } from 'react-bootstrap';
 import { postQuestionSubmission, putQuestionGrade, putQuestionGradeInstance, postPreviewQuestion, getQuestion } from '../APIInterfaces/BackendAPI/Requests/CourseRequests';
 import moment from 'moment';
 import { useCurrentProblemState } from '../Contexts/CurrentProblemState';
@@ -9,6 +9,7 @@ import { xRayVision } from '../Utilities/NakedPromise';
 import IframeResizer, { IFrameComponent } from 'iframe-resizer-react';
 import logger from '../Utilities/Logger';
 import BackendAPIError from '../APIInterfaces/BackendAPI/BackendAPIError';
+import useAlertState from '../Hooks/useAlertState';
 
 interface ProblemIframeProps {
     problem: ProblemObject;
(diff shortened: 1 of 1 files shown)


src/Assignments/ProblemIframe.tsx
```diff
@@ -364,7 +364,14 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
     return (
         <>
             { loading && <Spinner animation='border' role='status'><span className='sr-only'>Loading...</span></Spinner>}
-            {error && <div>{error}</div>}
+            {error &&
```
red-tt-bear, line 367: Why aren't we using the `useAlert` hook that we use everywhere else?
variant: danger

src/Assignments/ProblemIframe.tsx
```diff
@@ -332,7 +332,7 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
             updateSubmitActive();
         } else {
             if (renderedHTML !== '') {
-                logger.error('this problem has no problemMainForm'); // should NEVER happen when renderedHTML is non-empty
+                logger.error('this problem has no problemMainForm', renderedHTML); // should NEVER happen when renderedHTML is non-empty
```
red-tt-bear, line 335: I'm going to be a hypocrite here because I think I have done this too

According to winston logger the variable arguments are "meta", which if you use it with an error object is always fine

However if you do logger.error(string,string) on the backend the second string is omitted (I know we are on the frontend here)

When I set up the logger on the frontend I got it to work where it would behave the same as console.error since we were migrating over, however, we should probably avoid it 🤷‍♂️ 

It would just be 
```
logger.error(`this problem has no problemMainForm ${renderedHTML}`)
```

src/Assignments/ProblemIframe.tsx
```diff
@@ -364,7 +384,7 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
     return (
         <>
             { loading && <Spinner animation='border' role='status'><span className='sr-only'>Loading...</span></Spinner>}
-            {error && <div>{error}</div>}
+            <Alert variant={alert.variant} show={alert.message.length > 0}>{alert.message} -- Please refresh your page.</Alert>
```
PanopticaRising, line 387: Perfect!

PanopticaRising (approved):

src/Assignments/ProblemIframe.tsx
```diff
@@ -222,7 +228,10 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
 
                 if(_.isNil(iframeRef?.current)) {
                     logger.error('Hijacker: Could not find the iframe ref');
-                    setError('An error occurred');
+                    setAlert({
+                        variant: 'danger',
+                        message: 'There was an error with the problem iframe.'
```
red-tt-bear, line 233: This error message is very developery
A student is likely going to be offput by `iframe`

A better error message might look like:
`There was an error rendering the problem.`

PR #286, opened 2020-11-11
RED-107 Typed verification

red-tt-bear:
Displayed the error message

### DIFF
--- a/src/APIInterfaces/BackendAPI/ResponseTypes/UserResponseTypes.ts
+++ b/src/APIInterfaces/BackendAPI/ResponseTypes/UserResponseTypes.ts
@@ -23,3 +23,6 @@ type RegisterUser = {
     verificationBypass: boolean;
 };
 export type RegisterUserResponse = AxiosResponse<BackendAPIResponse<RegisterUser>>;
+
+type GetVerification = unknown;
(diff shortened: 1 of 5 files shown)


src/APIInterfaces/BackendAPI/ResponseTypes/UserResponseTypes.ts
```diff
@@ -22,4 +22,7 @@ export type GetUserResponse = AxiosResponse<GetUser>;
 type RegisterUser = {
     verificationBypass: boolean;
 };
-export type RegisterUserResponse = AxiosResponse<BackendAPIResponse<RegisterUser>>;
\ No newline at end of file
+export type RegisterUserResponse = AxiosResponse<BackendAPIResponse<RegisterUser>>;
+
+type GetVerification = unknown;
```
PanopticaRising, line 27: Why is this `unknown`? If we're not sending anything back, shouldn't it be `never`?

PanopticaRising (approved):

src/APIInterfaces/BackendAPI/ResponseTypes/UserResponseTypes.ts
red-tt-bear: It returns something we just don't use it so I didn't type it
I don't know what advantage never would have

doombot-exe (approved):

PR #292, opened 2020-11-17
Feature/red 282 grading page live look

doombot-exe:


src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +41,260 @@ const useStyles = makeStyles((theme) => ({
...
-        partialCreditBestScore: grade.partialCreditBestScore,
-        effectiveScore: grade.effectiveScore,
-        workbookId,
-    });
+        // workbookList?: WorkbookOption[];
+    }>({});
+    const displayCurrentScore = useRef<string | null>(null);
+    // fetch grade first
+    if (_.isNil(displayCurrentScore.current) && !_.isNil(grade)) {
+        displayCurrentScore.current = (grade?.effectiveScore * 100).toFixed(1);
```
red-tt-bear, line 68: toPercentString would be better

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +41,260 @@ const useStyles = makeStyles((theme) => ({
...
-                        // there should only be one studentGradeId in this case...
-                        (map[wb.studentGradeId] || (map[wb.studentGradeId] = [])).push(parseInt(id, 10));
-                        if (wb.studentGradeId !== grade.id) logger.error(`Workbooks for grade ${grade.id} are out of sync with the corresponding grade object.`);
-                    }
-                    return map;
-                }, {} as { [k: number]: number[] });
+    const fetchGrade = async () => {
+        setGrade(null);
+        if (!_.isNil(selected.user) && !_.isNil(selected.problem)) {
+            const res = await getQuestionGrade({
```
red-tt-bear, line 74: We should try catch and set error message

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +41,260 @@ const useStyles = makeStyles((theme) => ({
...
-    }>({
-        legalScore: grade.legalScore,
-        overallBestScore: grade.overallBestScore,
-        partialCreditBestScore: grade.partialCreditBestScore,
-        effectiveScore: grade.effectiveScore,
-        workbookId,
-    });
+        // workbookList?: WorkbookOption[];
+    }>({});
+    const displayCurrentScore = useRef<string | null>(null);
```
red-tt-bear, line 65: Looks like this was already there, but why do we need to useRef here?

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +41,260 @@ const useStyles = makeStyles((theme) => ({
...
+                studentGradeInstanceId,
+                attemptsCount: currentAttemptsCount,
+                averageScore: currentAverageScore,
+                versionMap: currentVersionMap,
+                // workbookList: currentWorkbookList,
+            });
         } else {
-            setSelected(selected => ({...selected, workbook: undefined})); // 
-            logger.debug(`GradeInfoHeader: No workbooks to select from or ${info.workbookId} cannot be found.`);
+            // this should not be a warning -- grade.id is initially null...
```
red-tt-bear, line 152: logger.debug seems appropriate here

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +41,260 @@ const useStyles = makeStyles((theme) => ({
...
+        logger.debug('GradeInfoHeader: setting new selected grades', newProblemState, newGradeInstance);
+        setSelected(selected => ({ ...selected, problemState: newProblemState, grade: currentGrade, gradeInstance: newGradeInstance }));
+    }, [info.workbookId, info.studentGradeId, info.studentGradeInstanceId]);
 
     const fetchTopicGrade = async () => {
         setTopicGrade(null);
-        if (_.isNil(topicId)) {
-            logger.debug('GradeInfoHeader topicId id is nil, skipping grades call');
+        if (_.isNil(topic)) {
+            logger.debug('GradeInfoHeader: topic id is nil, skipping grades call');
```
red-tt-bear, line 189: If we are skipping shouldn't there be a return?

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +41,260 @@ const useStyles = makeStyles((theme) => ({
...
-        const versionKeys = _.keys(vMap).map((v) => { return parseInt(v, 10); }).sort();
+    const onSuccess = (gradeOverride: Partial<StudentGrade>) => {
+        logger.debug('GradeInfoHeader: overriding grade', gradeOverride.effectiveScore);
+        const currentGrade = grade; 
+        if (!_.isNil(gradeOverride) &&
+            !_.isNil(gradeOverride.effectiveScore) &&
+            !_.isNil(currentGrade)
+        ) {
+            currentGrade.effectiveScore = gradeOverride.effectiveScore;
+            setGrade(currentGrade);
```
red-tt-bear, line 212: This shouldn't trigger a rerender since it is the same object

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +41,260 @@ const useStyles = makeStyles((theme) => ({
...
 
+    const versionSubList = (vMap: Record<number, Array<number>>, versionKey: number) => {
+        const attempts: WorkbookOption[] = [{ label: 'current', value: -1 }];
+        vMap[versionKey].sort().forEach((id, index) => {
+            attempts.push({ label: `Attempt #${index+1}`, value: id});
+        });
+        return attempts;
+    };
+
+    // const workbookList = (vMap: Record<number, Array<number>>, versioned: boolean) => {
```
red-tt-bear, line 233: Removing commented out code

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +41,260 @@ const useStyles = makeStyles((theme) => ({
...
+    //         vMap[key].sort().forEach((v, i) => {
+    //             const value = v;
+    //             const label = prefix + `Attempt #${(i + 1)}`;
+    //             options.push({ label, value });
+    //         });
+    //     });
+    //     return options;
+    // };
+
     type Props = {
```
red-tt-bear, line 248: This is old but we should follow our naming convention for props, that way we can move it to a separate file if we want or export it so that the type can be used elsewhere with the name being mindful

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +41,260 @@ const useStyles = makeStyles((theme) => ({
...
-        options: WorkbookOption[];
-        value: number;
+        versionMap: Record<number, Array<number>>;
+        versionKey?: number;
+        attemptKey: number;
         onChange: (s: typeof info) => void;
     };
 
     function WorkbookSelect<T extends KeyType>(props: Props) {
```
red-tt-bear, line 255: Again old but should this component be in another file

src/Courses/TopicGrades/GradingPage.tsx
```diff
@@ -61,168 +58,18 @@ export const TopicGradingPage: React.FC<TopicGradingPageProps> = () => {
...
-                .sortBy(['problemNumber'], ['asc'])
-                .value();
-            // currentProblems = _.map(currentProblems, (p) => {return new ProblemObject(p);});
-            setProblems(currentProblems);
-
-            const currentTopic = res.data.data.topic;
-            setTopic(currentTopic);
-        }
+        // fetchProblems is only called when topicId changes ()
+        const res = await getTopic({ id: topicId, includeQuestions: true });
```
red-tt-bear, line 63: Should have wrapped in try catch and setting error

src/Courses/TopicGrades/GradingPage.tsx
```diff
@@ -232,11 +79,10 @@ export const TopicGradingPage: React.FC<TopicGradingPageProps> = () => {
 
         if (!_.isEmpty(currentProblems)) {
             if (_.isNil(problemIdString)) {
-                const initialSelectedProblemId = _.sortBy(currentProblems, ['problemNumber'], ['asc'])[0].id;
-                initialSelectedProblem = currentProblemMap[initialSelectedProblemId] as ProblemObject;
+                initialSelectedProblem = currentProblems[0];
```
red-tt-bear, line 82: Could also use first

red-tt-bear (changes requested):
Error handling on requests

src/Courses/TopicGrades/GradeInfoHeader.tsx
doombot-exe: IDK, I thought you added this for showing the topic total score

src/Courses/TopicGrades/GradeInfoHeader.tsx
doombot-exe: not sure what you're saying - but this does update the displayed effective score when an override is made...

src/Courses/TopicGrades/GradeInfoHeader.tsx
red-tt-bear: I did because we needed to manage the parameter that came in vs the current state (show that you have a 75 and then an input that lets you make it 100)

But we are reusing that component right? or did we copy it?

src/Courses/TopicGrades/GradeInfoHeader.tsx
red-tt-bear: Gotcha, there is probably something else triggering the rerender

State only calls render when the reference changes, so for primitives like numbers it's whenever the value changes

However when you setState with an object and you set it to itself it does not trigger a rerender

That is why in a lot of places you'll see stuff like:
```
setTopic(new Topic(currentTopic))
```
or
```
setTopic({...topic})
```

Those are different ways of updating the reference

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -42,154 +46,268 @@ const useStyles = makeStyles((theme) => ({
...
+            versions.push({ label: `Version #${i+1}`, value: key});
         });
-        return options;
+        return versions;
     };
 
-    type Props = {
-        options: WorkbookOption[];
-        value: number;
+    const versionSubList = (vMap: Record<number, Array<number>>, versionKey: number) => {
```
PanopticaRising, line 249: I like these breakouts.

src/Courses/TopicGrades/GradingPage.tsx
```diff
@@ -25,248 +32,100 @@ export const TopicGradingPage: React.FC<TopicGradingPageProps> = () => {
...
+            const res = await getTopic({ id: topicId, includeQuestions: true });
+            const currentProblems = _(res.data.data.questions)
+                .map((p) => { return new ProblemObject(p); })
                 .sortBy(['problemNumber'], ['asc'])
                 .value();
             // currentProblems = _.map(currentProblems, (p) => {return new ProblemObject(p);});
             setProblems(currentProblems);
 
-            const currentTopic = res.data.data.topic;
+            const currentTopic = res.data.data as TopicObject;
```
PanopticaRising, line 68: I think you should use `new TopicObject` instead of casting, because any logic we'll have on the class won't run otherwise.

PanopticaRising:
```suggestion
            const currentTopic = new TopicObject(res.data.data);
```

PanopticaRising (approved):

src/Courses/TopicGrades/GradeInfoHeader.tsx
red-tt-bear: This one's on me...

src/Courses/TopicGrades/GradeInfoHeader.tsx
red-tt-bear: Maybe add a TODO to investigate in the future

red-tt-bear (approved):

PR #304, opened 2020-11-23
Added better README and a potential proxy rule

PanopticaRising:


README.md
```diff
@@ -1,44 +1,18 @@
-This project was bootstrapped with [Create React App](https://github.com/facebook/create-react-app).
+# Rederly Frontend
 
-## Available Scripts
+![Version](https://img.shields.io/github/v/release/rederly/frontend?style=plastic)
```
red-tt-bear, line 3: How is this getting it's status, is this hooked up to the github action?

src/setupProxy.js
```diff
@@ -10,4 +10,8 @@ module.exports =  function(app) {
     app.use(createProxyMiddleware('/webwork2_files', {
         target: 'http://localhost:3000'
     }));
+
+    app.use(createProxyMiddleware('/work', {
```
red-tt-bear, line 14: I think we should add dev configuration, possibly as environment variables if it doesn't affect the build

Not now but in the future we should think through howe we test the lambda in development

I think @Jhuanderson-Rederly said there is a way to run it locally, hopefully in docker

README.md
PanopticaRising: This is getting it from our Releases tab.

src/setupProxy.js
PanopticaRising: I just finished testing it and it didn't seem to work well with the redirect to stage, so I just removed it.

red-tt-bear (approved):

PR #310, opened 2020-11-25
Feature/red 287 Part 1 - Promisifying Img OnLoads

PanopticaRising:
This is a partial PR because it handles one of three cases that need to be awaited during the loading of the print page- img, iframes, and pdfs.


src/Contexts/PrintLoadingContext.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+export enum PrintLoadingActions {
+    ADD_PROMISE
+}
+
+const reducer = (state: Promise<unknown>[], action: PrintLoadingDispatch): Promise<unknown>[] => {
+    logger.debug('Adding a promise to the Loading context');
+    switch(action.type) {
+    case PrintLoadingActions.ADD_PROMISE:
+        action.payload.finally(()=>console.log('This embed finished loading.'));
```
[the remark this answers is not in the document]
red-tt-bear: Busted!!!

src/Hooks/useLoading.tsx
```diff
@@ -0,0 +1,15 @@
+import { useState } from "react";
```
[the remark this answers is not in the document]
red-tt-bear: Nice I didn't realize it did this!!!

.vscode/settings.json
```diff
@@ -1,3 +1,4 @@
 {
-    "editor.wordBasedSuggestions": false
+    "editor.wordBasedSuggestions": false,
+    "typescript.tsdk": "node_modules/typescript/lib"
```
red-tt-bear, line 3: What does this do?

red-tt-bear: Just making sure we are using the same version of typescript?

src/Assignments/ProblemIframe.tsx
```diff
@@ -348,7 +355,7 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
             // check that the submit url is accurate
             const submitUrl = problemForm.getAttribute('action');
             const checkId = submitUrl?.match(/\/backend-api\/courses\/question\/([0-9]+)\?/);
-            if (checkId && parseInt(checkId[1],10) !== problem.id) {
+            if (!checkId || parseInt(checkId[1],10) !== problem.id) {
```
red-tt-bear, line 358: I think we said we would revert this

src/Contexts/PrintLoadingContext.tsx
```diff
@@ -0,0 +1,90 @@
...
+    useEffect(()=>{
+        if (state.expected > 1 && state.arr.length >= state.expected && Promise.allSettled(state.arr)) {
+            logger.debug(`${state.expected} components have loaded, printing in two seconds.`);
+
+            setTimeout(() => {
+                dispatch({
+                    type: PrintLoadingActions.PRINT,
+                    expected: state.expected,
+                });
+            }, 10000);
```
red-tt-bear, line 77: I thought we didn't need the 10 second wait, I think I left this comment on an old pr but didn't see the resolution

src/Courses/TopicGrades/PrintEverything.tsx
```diff
@@ -6,52 +6,69 @@ import { ProblemObject } from '../CourseInterfaces';
...
 export const PrintEverything: React.FC<PrintEverythingProps> = () => {
-    let userId: number = 0;
-    let topicId: number = 0;
-    const qs = useQuery();
-    const workbookName = qs.get('workbookName');
+    const [gradeData, setGradeData] = useState<GetAllVersionAttachmentsResponse | null>(null);
+    const {dispatch} = usePrintLoadingContext();
     const params = useParams<{userId?: string, topicId?: string}>();
 
+    let userId: number = 0;
```
red-tt-bear, line 24: I don't like defaulting to 0
Also there is an error that checks if they are nan and says they must be in the url, but if they are not it will let them through

Instead you could set them to null and down there check that they are both a number, you will accomplish the same thing without having a default value that could break the logic

src/Utilities/promiseOnLoad.ts
```diff
@@ -0,0 +1,20 @@
...
+        });
+        elementRef.addEventListener('abort', () => {
+            reject();
+        });
+    }
+).catch(()=>{
+    logger.warn('An element failed to load and rejected promiseOnLoad.');
+});
+
+export default promiseOnLoad;
```
red-tt-bear, line 20: Missing newline

.vscode/settings.json
PanopticaRising: I think this forces VS code to use the version we're using, rather than a different version for type hinting.

src/Contexts/PrintLoadingContext.tsx
PanopticaRising: No, I responded to your comment; we still need some delay to account for all possible loading delays.

red-tt-bear (approved):

PR #311, opened 2020-11-27
Feature/red 287 pdf performance

PanopticaRising:


src/Assignments/ProblemIframe.tsx
```diff
@@ -348,7 +355,7 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
             // check that the submit url is accurate
             const submitUrl = problemForm.getAttribute('action');
             const checkId = submitUrl?.match(/\/backend-api\/courses\/question\/([0-9]+)\?/);
-            if (checkId && parseInt(checkId[1],10) !== problem.id) {
+            if (!checkId || parseInt(checkId[1],10) !== problem.id) {
```
red-tt-bear, line 358: @doombot-exe this was the proposed change in slack

src/Assignments/ProblemIframe.tsx
```diff
@@ -382,11 +389,17 @@ export const ProblemIframe: React.FC<ProblemIframeProps> = ({
...
-            }); 
-            await Promise.all(promises);       
+            });
+            await Promise.all(promises);
         }
 
-        setLoading(false);
+        setRenderedHTML((renderedHTML: string): string => {
+            if (renderedHTML !== '') {
+                iframeDoc?.dispatchEvent(new Event('Rederly Loaded'));
```
red-tt-bear, line 398: I gave this string as an example, I'm fine with it but I don't know if matches conventions (with the caps and the space)

src/Utilities/promiseOnLoad.ts
```diff
@@ -17,4 +18,10 @@ export const promiseOnLoad = (elementRef: HTMLElement) => new Promise(
     logger.warn('An element failed to load and rejected promiseOnLoad.');
 });
 
+export const promiseIframeOnLoad = (problemIframeRef: React.FC) => new Promise(
```
red-tt-bear, line 21: What is this, we should remove it

red-tt-bear (changes requested):
Looks good, just need to resolve eslint issues
I'd probably be fine with them but the debugging promise I think has to go

src/Assignments/ProblemIframe.tsx
red-tt-bear: @doombot-exe said he is against this in cases like static problem where there is no url period
I still think it should be here, however I'm fine with doing:
```
((submitUrl && !checkId) || parseInt(checkId[1],10) !== problem.id))
```

drdrew42: it's not that the submitUrl won't exist though... I'm only saying that it's not worth checking the problemId if the submitUrl doesn't match the RegEx. We _will_ have issues where the iframe renders content that doesn't belong to a "problem" -- I'm thinking of problem editing or elsewhere that has previews without being "formally" a topic-problem.

red-tt-bear (approved):

PR #314, opened 2020-12-01
fix setLoading state

doombot-exe:
this is my bad -- I added the loading state and didn't account for all situations to setLoading(false)


red-tt-bear (approved):
Let's get rid of the info logs

PanopticaRising (approved):

PR #317, opened 2020-12-02
Feature/red 287 expectation tracking promises

PanopticaRising:


src/Contexts/PrintLoadingContext.tsx
```diff
@@ -1,40 +1,85 @@
...
+    RESET_EXPECTED_COUNT,
+    PRINT,
 }
 
-const reducer = (state: Promise<unknown>[], action: PrintLoadingDispatch): Promise<unknown>[] => {
-    logger.debug('Adding a promise to the Loading context');
+const reducer = (state: PrintLoadingState, action: PrintLoadingDispatch): PrintLoadingState => {
     switch(action.type) {
+    case PrintLoadingActions.RESET_EXPECTED_COUNT:
+        logger.debug('Printing state reset.');
```
red-tt-bear, line 39: We might want to prepend some sort of tag, not a big deal though since I would like to be able to add that to the logger as awhole

src/Contexts/PrintLoadingContext.tsx
```diff
@@ -1,40 +1,85 @@
...
+    useEffect(()=>{
+        if (state.expected > 1 && state.arr.length >= state.expected && Promise.allSettled(state.arr)) {
+            logger.debug(`${state.expected} components have loaded, printing in two seconds.`);
+
+            setTimeout(() => {
+                dispatch({
+                    type: PrintLoadingActions.PRINT,
+                    expected: state.expected,
+                });
+            }, 10000);
```
red-tt-bear, line 77: I thought you were working on getting rid of the 10 seconds wait? Is that coming later

PanopticaRising: No, the 10 seconds is still going to stay to catch things we can't guarantee are tracked properly right now. These PRs address the 90% case where 10 seconds is too little time.

red-tt-bear (approved):

PR #318, opened 2020-12-02
Feature/red 235 add partialcreditbestscore

PanopticaRising:


doombot-exe (approved):

src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -310,13 +310,15 @@ export const GradeInfoHeader: React.FC<GradeInfoHeaderProps> = ({
                 <h4>Statistics</h4>
                 Number of attempts: <strong>{info.attemptsCount}</strong><br />
                 Best overall score: <strong>{info.overallBestScore?.toPercentString()}</strong><br />
+                {/* Should check !moment(topic.endDate).isSame(topic.deadDate) for topic and overrides before showing this? */}
+                Best score (including Partial Credit): <strong>{info.partialCreditBestScore?.toPercentString()}</strong><br />
```
red-tt-bear, line 314: We need to be careful, this is unclear what this difference is to the user

I think this should simply say:
`System Score`

As in the score you would have gotten from the system without professor intervention, I think adding more details makes it confusing

red-tt-bear: Also it might have been unclear from the ticket but this is what I call the `Topic Grading Page`
We already had a grading page and it was the grids

src/Utilities/NakedPromise.ts
```diff
@@ -3,7 +3,7 @@ import logger from './Logger';
 export class NakedPromise<T> {
     public promise: Promise<T>;
     public reject!: (() => void);
-    public resolve!: (() => void);
+    public resolve!: ((value: T | PromiseLike<T>) => void);
```
red-tt-bear, line 6: What benefit does this have?

PanopticaRising: Typescript doesn't throw errors 😂

red-tt-bear (approved):

PR #319, opened 2020-12-02
Added Modernizr support, scripts to build it, typing, and a check to …

PanopticaRising:
…use with webp images, and a new png fallback.


package.json
```diff
@@ -62,10 +63,13 @@
...
     "test": "react-scripts test",
     "eject": "react-scripts eject",
+    "prestart": "npm run modernizr",
+    "prebuild": "npm run modernizr",
     "build": "react-scripts build",
     "rebuild_renderer": "(cd ../renderer-docker && ./rebuild_docker.sh)",
     "lint": "eslint src/ --ext ts,tsx",
-    "postbuild": "node postbuild.js"
+    "postbuild": "node postbuild.js",
+    "modernizr": "npx modernizr -c modernizr-config.json -d public/modernizr.min.js"
```
red-tt-bear, line 72: Still not a fan of npx, if someone removed  from package.json not realizing it is used then instead of failing it would just use a different version

src/NavWrapper/NavWrapper.tsx
```diff
@@ -82,7 +83,12 @@ export const NavWrapper: React.FC<NavWrapperProps> = () => {
             <Navbar role='navigation' variant='dark' bg='dark' className="toolbar mr-auto">
                 <NavbarBrand as={Link} to="/common/courses">
                     <img
-                        src="/rederly-logo-offwhite.webp"
+                        src={
+                            // Fair warning, don't === this, it's not a real boolean.
```
red-tt-bear, line 87: Thanks 😃

red-tt-bear (approved):

package.json
PanopticaRising: Nothing to do with being a fan of npx, it just has no purpose being inside the package.json, I just copied it from the commands I was running outside it.

red-tt-bear (approved):

PR #327, opened 2020-12-07
Feature/red 235 add partialcreditbestscore

PanopticaRising:


src/Courses/CourseDetailsTabs/GradeTable.tsx
```diff
@@ -16,15 +17,11 @@ export const GradeTable: React.FC<GradeTableProps> = ({courseName, grades, onRow
...
+        ...(_.isNil(obj.average) ? undefined : {average: obj.average.toPercentString()}),
+        ...(_.isNil(obj.effectiveScore) ? undefined : {effectiveScore: obj.effectiveScore.toPercentString()}),
+        ...(_.isNil(obj.systemScore) ? undefined : {systemScore: obj.systemScore.toPercentString()}),
     }));
     const headers = _(safeGrades[0]).keys().filter(n => n !== 'id').value();
-    if(headers.indexOf('average') >= 0) {
-        // Would include this in above mapping, however using ternary operated resulted in an empty column in questions
-        safeGrades = safeGrades.map(obj => ({
-            ...obj,
-            average: _.isNil(obj.average) ? '--' : `${(obj.average * 100).toFixed(1)}%`
```
red-tt-bear, line 25: What happened to the `--`

red-tt-bear: This happens if the weight is 0 therefore it would be divide by zero and it returns null

red-tt-bear (changes requested):
What happens when the result of the query is null (more details on the -- comment

src/Courses/CourseDetailsTabs/GradeTable.tsx
PanopticaRising: "I mean optional" - @red-tt-bear 

I need to check both if the fields are present in the headers AND if they're null.

red-tt-bear (approved):

doombot-exe (approved):

PR #336, opened 2020-12-10
Added tabs to URL with fallback support.

PanopticaRising:
Also set Bootstrap to not mount the tabs below until they're navigated to, which prevents frontloading all our API calls.

### DIFF
--- a/src/Courses/CourseDetailsPage.tsx
+++ b/src/Courses/CourseDetailsPage.tsx
@@ -1,13 +1,22 @@
 import React, { useState } from 'react';
 import { Container, Tabs, Tab } from 'react-bootstrap';
-import EnrollmentsTab from './CourseDetailsTabs/EnrollmentsTab';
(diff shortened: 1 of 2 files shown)


src/Courses/CourseDetailsPage.tsx
```diff
@@ -46,37 +59,62 @@ export const CourseDetailsPage: React.FC<CourseDetailsPageProps> = () => {
...
         <Container>
-            <Provider value={course}>    
-                <Tabs 
-                    activeKey={activeTab} 
-                    defaultActiveKey={CourseDetailsTabs.DETAILS} 
-                    id="course-details-tabs" 
+            <Provider value={course}>
+                <Tabs
+                    activeKey={activeTab}
+                    defaultActiveKey={tab ?? CourseDetailsTabs.DETAILS}
```
red-tt-bear, line 65: Why not default this up top?

red-tt-bear (approved):

PR #342, opened 2020-12-17
Fixed printing early bugs related to rendering, added new library to help debug problems.

PanopticaRising:


src/Contexts/PrintLoadingContext.tsx
```diff
@@ -33,50 +33,64 @@ export enum PrintLoadingActions {
...
-    }
-    return state;
-};
-
 export const usePrintLoadingContext = () => React.useContext(PrintLoadingContext);
 
 export const PrintLoadingProvider: React.FC<Props>  = ({ children }) => {
-    const [state, dispatch] = useReducer(reducer, {expected: 1, arr: []});
+    const timeout = React.useRef<NodeJS.Timeout | null>(null);
+    const state: PrintLoadingState = {expected: 0, arr: []};
```
red-tt-bear, line 40: Should this be a use ref?
Also we should probably rename from state since it is confusing

src/Contexts/PrintLoadingContext.tsx
```diff
@@ -33,50 +33,64 @@ export enum PrintLoadingActions {
...
 
-            setTimeout(() => {
-                dispatch({
-                    type: PrintLoadingActions.PRINT,
-                    expected: state.expected,
-                });
-            }, 10000);
+            // If a promise was added and there's already a timeout waiting to be evaluated, remove it.
+            if (!_.isNull(timeout.current)) {
+                clearTimeout(timeout.current);
```
red-tt-bear, line 65: After clearing the timeout shouldn't you set it to null?
Clearing a timeout multiple times doesn't have any adverse effects I don't think but just to have a source of truth

src/wdyr.ts
```diff
@@ -0,0 +1,11 @@
+/// <reference types="@welldone-software/why-did-you-render" />
+import React from 'react';
+
+// This is useful for debugging rerendering problems.
+// https://github.com/welldone-software/why-did-you-render
+if (process.env.NODE_ENV === 'development' && process.env.REACT_APP_ENABLE_WDYR === 'true') {
```
red-tt-bear, line 6: What happens to env variables after you deploy?

red-tt-bear (approved):

src/wdyr.ts
PanopticaRising: They get inlined.

PR #344, opened 2020-12-17
Added a useEffect hook to update the URL every time the selected obje…

PanopticaRising:
…cts change.

### DIFF
--- a/src/Courses/TopicGrades/GradingPage.tsx
+++ b/src/Courses/TopicGrades/GradingPage.tsx
@@ -2,7 +2,7 @@ import { Grid, Button, Snackbar } from '@material-ui/core';
 import { Alert as MUIAlert } from '@material-ui/lab';
 import React, { useEffect, useState } from 'react';
 import _ from 'lodash';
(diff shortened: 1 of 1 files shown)


src/Courses/TopicGrades/GradingPage.tsx
```diff
@@ -38,6 +39,17 @@ export const TopicGradingPage: React.FC<TopicGradingPageProps> = () => {
         grade?: StudentGrade,
         gradeInstance?: StudentGradeInstance,
     }>({});
+    const { url } = useRouteMatch();
+    const history = useHistory();
+
+    useEffect(() => {
+        const queryString = qs.stringify(_({
```
red-tt-bear, line 46: Does this trigger a re render from react routers?

red-tt-bear (approved):

src/Courses/TopicGrades/GradingPage.tsx
PanopticaRising: I haven't looked specifically if it causes rerenders higher up, but the page exhibits no stuttering.

PR #349, opened 2020-12-18
Feature/red 10 tarball import

red-tt-bear:


src/Courses/CourseDetailsTabs/TopicsTab.tsx
```diff
@@ -114,6 +117,12 @@ export const TopicsTab: React.FC<TopicsTabProps> = ({ course, setCourse }) => {
         createTopic(courseUnitContentId);
     };
 
+    const setUnitInCourse = (unit: UnitObject) => {
```
PanopticaRising, line 120: I would call this `addUnitToCourse`. The `set` confused me, when really all it's doing is pushing a unit onto the stack.

src/Courses/CourseDetailsTabs/TopicsTab.tsx
```diff
@@ -417,13 +426,49 @@ export const TopicsTab: React.FC<TopicsTabProps> = ({ course, setCourse }) => {
...
+                                            switch (event.status) {
+                                            case 'error':
+                                                setError(event.data);
+                                                break;
+                                            case 'success':
+                                                setUnitInCourse(event.data);
+                                                break;
+                                            case 'loading':
+                                                setLoading(true);
+                                                break;
```
PanopticaRising, line 454: This is probably a good example of where one state is better than multiple, and a reducer might help.

As it is, the `switch` doesn't add much value when it'd be more clearly expressed as `if/else`s.

PanopticaRising (approved):

doombot-exe (approved):

PR #357, opened 2020-12-24
RED-321/local storage and read config json

doombot-exe:


src/Login/LoginForm.tsx
```diff
@@ -77,6 +78,29 @@ export const LoginForm: React.FC<LoginFormProps> = () => {
...
+            const res = await AxiosRequest.get('/users/status');
+            const data = res.data.data as { userPaidUntil: Date, universityPaidUntil: Date };
+            const userPaidMoment = data.userPaidUntil.toMoment();
+            const universityPaidMoment = data.universityPaidUntil.toMoment();
+
+            let paidUntil: moment.Moment | undefined;
+            let accountType: string | undefined;
+            if (userPaidMoment.isAfter(universityPaidMoment)) {
+                paidUntil = userPaidMoment;
+                accountType = 'INDIVIDUAL';
```
PanopticaRising, line 92: Should use the enum here.

src/Login/LoginForm.tsx
```diff
@@ -77,6 +78,29 @@ export const LoginForm: React.FC<LoginFormProps> = () => {
                 setLoginAlertMsg({ message: err.message, variant: 'danger' });
             }
         }
+
+        try {
+            const res = await AxiosRequest.get('/users/status');
```
red-tt-bear, line 83: Why are we doing this on the login form?

At one point we talked about returning this in the login response but we decided to make a separate endpoint and use it as needed since things aren't ready to be reactive but that would keep it the most relevant

src/Login/LoginForm.tsx
```diff
@@ -77,6 +78,29 @@ export const LoginForm: React.FC<LoginFormProps> = () => {
...
+            const userPaidMoment = data.userPaidUntil.toMoment();
+            const universityPaidMoment = data.universityPaidUntil.toMoment();
+
+            let paidUntil: moment.Moment | undefined;
+            let accountType: AccountType | undefined;
+            if (userPaidMoment.isAfter(universityPaidMoment)) {
+                paidUntil = userPaidMoment;
+                accountType = AccountType.INDIVIDUAL;
+            } else {
+                paidUntil = universityPaidMoment;
```
red-tt-bear, line 94: The paid until field on the user is the source of truth, in theory, it should always be equal to or after the university's paid until but I don't want to impose that restriction

If we ever wanted to block a professor for a semester for whatever reason we could reduce their paid until and until the university paid their account would be blocked

src/index.tsx
```diff
@@ -7,6 +7,24 @@ import { Router } from './Router';
...
+}
+
+declare global {
+    interface Window {
+        rederlyConfig?: Promise<RederlyConfig>;
+    }
+}
+
+const getConfig = async () => {
+    window.rederlyConfig = axios.get<RederlyConfig>(`/config.json?cache_bust=${new Date().getTime()}`).
```
red-tt-bear, line 23: We should catch this and log an error

red-tt-bear (changes requested):
Various comments

src/index.tsx
```diff
@@ -7,6 +7,24 @@ import { Router } from './Router';
...
+        then(resp => resp.data);
```
red-tt-bear, line 24: .catch(err => {
    logger.error('Failed to fetch frontend configuration', e);
    return null;
});

src/index.tsx
```diff
@@ -7,6 +7,24 @@ import { Router } from './Router';
...
+};
+
+getConfig();
```
red-tt-bear, line 27: const getConfig = () => axios.get(...);
window.rederlyConfig = getConfig();

PanopticaRising (approved):

src/index.tsx
```diff
@@ -7,6 +7,28 @@ import { Router } from './Router';
...
+    paymentURL: string;
+}
+
+declare global {
+    interface Window {
+        rederlyConfig?: Promise<RederlyConfig | null>;
+    }
+}
+
+const getConfig = () => axios.get<RederlyConfig>(`/config.json?cache_bust=${new Date().getTime()}`).
```
red-tt-bear, line 23: NIT: it is common practice to put the `.` on the newline
```Javascript
Promise.resolve()
.then()
.catch()
```

src/Account/AccountDetailsPage.tsx
```diff
@@ -18,8 +27,17 @@ export const AccountDetailsPage: React.FC<AccountDetailsPageProps> = () => {
                     {_.chain(userName).words().map(w => w[0]).join('').value()}
                 </Avatar>
             </Grid>
-            <Grid container item xs={12} justify='center'>
+            <Grid container item xs={12} justify='center' direction='column' alignItems='center' spacing={5}>
                 <TextField id="user-name" label="Name" value={userName} disabled />
+                <Grid item>
```
red-tt-bear, line 32: This should only be visible for professors

red-tt-bear (changes requested):
. on newline for chaining
Move get status call to the course details page to keep it relevant
Hide paid until for students

src/Account/AccountDetailsPage.tsx
```diff
@@ -16,6 +19,28 @@ export const AccountDetailsPage: React.FC<AccountDetailsPageProps> = () => {
...
+                if (userPaidMoment.isAfter(universityPaidMoment)) {
+                    accountType = AccountType.INDIVIDUAL;
+                } else if (userPaidMoment.isSame(universityPaidMoment)) {
+                    accountType = AccountType.INSTITUTIONAL;
+                } else {
+                    accountType = AccountType.DISABLED;
+                }
+
+                account.paidUntil = paidUntil.toDate();
+                account.accountOwner = accountType;
```
red-tt-bear, line 40: Need to use state here so it will be reactive

red-tt-bear (approved):

PR #359, opened 2020-12-29
Feature/add edit mode to url

PanopticaRising:


src/Hooks/useQuerystringHelper.ts
```diff
@@ -56,4 +68,25 @@ export const useQuerystringHelper = () => {
     return {getQuerystring, updateRoute};
 };
 
+// https://stackoverflow.com/questions/8648892/how-to-convert-url-parameters-to-a-javascript-object
+// Functions like _.fromPairs and Object.fromEntries do not handle array cases, and instead overwrite repeated keys.
+// for..of won't work with Iterators unless we use --downleveliteration or target >es5
+const paramsToObject = (params: URLSearchParams) => {
```
red-tt-bear, line 74: I wonder if query string parse would do the trick

red-tt-bear: Seems to
```Javscript
> require('querystring').parse('a=1&a=2')
[Object: null prototype] { a: [ '1', '2' ] }
```

red-tt-bear (approved):
Looks good but we could simplify

src/Hooks/useQuerystringHelper.ts
PanopticaRising: Oh, cool, that works.

doombot-exe (approved):

PR #366, opened 2020-12-31
Feature/red 10 tarball import

red-tt-bear:


src/Courses/CourseDetailsTabs/TopicsTab.tsx
```diff
@@ -367,18 +437,50 @@ export const TopicsTab: React.FC<TopicsTabProps> = ({ course, setCourse }) => {
...
+                                                    });
+                                                } else {
+                                                    setAlert({
+                                                        variant: 'danger',
+                                                        message: event.data.message
+                                                    });
+                                                }
                                                 break;
-                                            case 'success':
+                                            case 'success': {
```
PanopticaRising, line 470: Do we not set a success modal?

src/Courses/CourseDetailsTabs/TopicsTab.tsx
```diff
@@ -347,7 +417,7 @@ export const TopicsTab: React.FC<TopicsTabProps> = ({ course, setCourse }) => {
                 headerContent={<h5>Confirm delete</h5>}
                 bodyContent={`Are you sure you want to remove ${confirmationParamters.identifierText}?`}
             />
-            {error && <Alert variant="danger">{error.message}</Alert>}
+            <Alert variant={alert.variant} show={Boolean(alert.message)}>{alert.message}</Alert>
```
PanopticaRising, line 420: nit: You could've pulled in the MUI Alert, which I think is slightly nicer.

PanopticaRising: It doesn't work with useAlert, but we have a useMUIAlert in the same file.

PanopticaRising (approved):

src/Courses/CourseDetailsTabs/TopicsTab.tsx
red-tt-bear: No because the unit comes up, this matches the existing add unit behavior

src/Courses/CourseDetailsTabs/TopicsTab.tsx
red-tt-bear: Yeah I was already making big changes and didn't want to rock the boat (plus it was 2am 🤣 )

doombot-exe (approved):

PR #370, opened 2021-01-02
RED-10 Added asset upload

red-tt-bear:
Fixed bug with error handling for archive uploads
Added asset upload to the editor

### DIFF
--- a/src/Assignments/ProblemEditorAssetUploadButton.tsx
+++ b/src/Assignments/ProblemEditorAssetUploadButton.tsx
@@ -1,0 +1,229 @@
+import React, { useState, useCallback, useEffect, ChangeEvent } from 'react';
+import { Box, TextField, Button, InputAdornment } from '@material-ui/core';
+// import { Button } from 'react-bootstrap';
+import _ from 'lodash';
+import { useDropzone } from 'react-dropzone';
+import { FaFileUpload } from 'react-icons/fa';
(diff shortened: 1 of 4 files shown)


PanopticaRising (approved):

src/Assignments/ProblemEditorAssetUploadButton.tsx
```diff
@@ -0,0 +1,229 @@
...
+                                            opacity: 0.9,
+                                        }}
+                                        >
+                                            <div style={{ position: 'relative', margin: '0 auto', top: '15%', fontSize: '1em' }}>
+                                                <FaFileUpload style={{ position: 'relative', margin: '0 auto', top: '15%', display: 'block', fontSize: '1em' }} />
+                                            </div>
+                                        </div>
+                                    )}
+                                </div>
+                            </Box> */}
```
doombot-exe, line 196: this seems to be duplicate code of 88-122?

doombot-exe (approved):

src/Assignments/ProblemEditorAssetUploadButton.tsx
red-tt-bear: It does seem that way but the idea is that you can change the item that you want to upload once in the modal, the problem I was facing was that the dropzone misbehaved on the first button when used twice

PR #373, opened 2021-01-06
SMA proof of concept

doombot-exe:


red-tt-bear (approved):

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -427,9 +482,9 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
         return <Spinner animation='border' role='status'><span className='sr-only'>Loading...</span></Spinner>;
     }
 
-    if (error) {
-        return <div>{error}</div>;
-    }
+    // if (error) {
```
red-tt-bear: Will have to discuss but I don't think this makes sense

On updates where the status is off to the side the alerts make sense

but for the problem, if it doesn't render you can't render anything other than the error

PR #377, opened 2021-01-07
Feature/red 322 show me another

doombot-exe:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "webwork-react-ts",
-  "version": "1.7.4",
+  "version": "1.7.4-demo4",
```
PanopticaRising, line 3: Are we committing this to dev?

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -44,13 +45,27 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
     const [modalLoading, setModalLoading] = useState<boolean>(false);
     const [selectedProblemId, setSelectedProblemId] = useState<number | null>(null);
     const [loading, setLoading] = useState(true);
-    const [error, setError] = useState('');
+    const [alert, setAlert] = useAlertState();
```
PanopticaRising, line 48: Since you changed this, you should consider using `useMUIAlertState` and use the Material UI version of Alert.

PanopticaRising: The API is similar, but MUI's `severity` is the equivalent of BS's `variant`.

src/Assignments/SimpleProblemPage.tsx
```diff
@@ -44,13 +45,27 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
...
     const [confirmationParameters, setConfirmationParameters] = useState<ConfirmationModalProps>(DEFAULT_CONFIRMATION_PARAMETERS);
     const [openDrawer, setOpenDrawer] = useState<boolean>(false);
     const {course, users} = useCourseContext();
 
+    const resetAlert = (): void => {
+        setAlert({
```
PanopticaRising, line 54: I like this, but we should statically store the object (with a `useRef` or so) this is using so that we avoid unnecessary rerenders.

PanopticaRising (approved):

PanopticaRising (approved):

red-tt-bear (approved):

PR #407, opened 2021-01-28
Used class styles to prevent overwriting. There's still a jump when w…

PanopticaRising:
…e drag that we should try to fix.


PanopticaRising:
I only lightly tested this, but I can drag a problem with the red error highlighting and reorder it.

red-tt-bear (approved):

doombot-exe (approved):

PR #416, opened 2021-02-02
remove outdated useEffect trigger

doombot-exe:
minor correction - fixes unnecessary re-render on first currentProblemState update.
useEffect now watches specific parameters of the problem object rather than the whole problem


doombot-exe:
now the 'saved' state can update any problems marked 'partial' or 'incorrect' -- useful for exams. 'Complete' problems will never be overriden with 'saved' state.

we might consider implications for standard problem sets - does a 'saved' icon give the wrong impression?

PanopticaRising (approved):

red-tt-bear (approved):

PR #419, opened 2021-02-03
Created new subfolder for tab and moved TopicsTab components into it.

PanopticaRising:


red-tt-bear (approved):

src/Courses/CourseDetailsTabs/ActiveTopics.tsx
```diff
@@ -11,42 +11,21 @@ interface ActiveTopicsProps {
...
-                        _.assign(topic, override);
-                    }
-                    return topic;
-                });
-            }
-
-            setActiveTopics(activeTopics);
-        })();
-    }, [course.id]);
+    const activeTopics = course.units.reduce<Array<TopicObject>>((accum, unit) => {
```
red-tt-bear, line 17: I wonder if you could have done this more simply with lodash flatten method, either way looks good

red-tt-bear: Should we `useMemo` here?

red-tt-bear: > I wonder if you could have done this more simply with lodash flatten method, either way looks good
I guess flatten doesn't make sense since we are also filtering

PanopticaRising: We'd have to add every topic to the dependency array, and I'm not sure if we can do that dynamically.

I see this as more of a holdover until we add SWR.

red-tt-bear (approved):

red-tt-bear (approved):

PR #424, opened 2021-02-05
Fixed color for disabled email state.

PanopticaRising:


src/Courses/CourseDetailsTabs/EmailComponentWrapper.tsx
```diff
@@ -58,12 +58,21 @@ export const EmailComponentWrapper: React.FC<EmailComponentWrapperProps> = ({ us
         icon: function IconWrapper() {
             return <span><Email style={{color: '#007bff'}}/> Email</span>;
         },
-        // isFreeAction: true,
         tooltip: 'Email selected students',
         onClick: () => setShowModal(true),
         position: 'toolbarOnSelect'
     };
 
+    const disabledEmailProfessorButtonOptions: Action<UserObject> = {
```
red-tt-bear, line 66: Why did you remove the spread from the above?
Only the changes to the disabled state should be defined

This way if we have a change in verbiage or icon it is just in one place

red-tt-bear (changes requested):
https://github.com/rederly/frontend/pull/424/files#r571198977

red-tt-bear (approved):

red-tt-bear (approved):

PR #429, opened 2021-02-10
Feature/add query params to simple problem page

PanopticaRising:


src/Assignments/SimpleProblemPage.tsx
```diff
@@ -43,13 +44,16 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
     };
 
     const params = useParams<SimpleProblemPageLocationParams>();
+    const {getCurrentQueryStrings, updateRoute} = useQuerystringHelper();
+    const urlProblemIdQS = getCurrentQueryStrings()?.['problemId'];
```
red-tt-bear, line 48: Why do you need the brackets and quotes?

PanopticaRising: That's the way getCurrentQueryStrings is structured. Everything in the URL is a string, and may or may not be there.

red-tt-bear (approved):

PR #435, opened 2021-02-16
RED-486 Fixed bug that blocked email until page refresh

red-tt-bear:
Email students now clears all state before bowing out

With this I added a state machine to make it a little bit cleaner:
The entire form is only enabled when ready, not while loading or after
submission
A loading spinner shows while the email is sending

This prevents profs from accidentally making the requests multiple times
really quickly

### DIFF
--- a/src/Courses/CourseDetailsTabs/EmailModal.tsx
+++ b/src/Courses/CourseDetailsTabs/EmailModal.tsx
@@ -1,6 +1,6 @@
 import _ from 'lodash';
 import React, { useState } from 'react';
-import { Modal, ModalTitle, ModalBody, Form, Button, ModalFooter, FormGroup, FormControl, FormLabel, Alert } from 'react-bootstrap';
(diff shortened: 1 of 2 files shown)


src/Courses/CourseDetailsTabs/EmailModal.tsx
```diff
@@ -12,62 +12,94 @@ interface EmailModalProps {
...
-export const EmailModal: React.FC<EmailModalProps> = ({users, show, setClose}) => {
+export const EmailModal: React.FC<EmailModalProps> = ({users, show, setClose: setCloseProp}) => {
     const [subject, setSubject] = useState('');
     const [body, setBody] = useState('');
     const [{message: sendEmailRespMsg, variant: sendEmailRespAlertType}, setSendEmailRespMsg] = useAlertState();
+    const [modalState, setModalState] = useState(EmailModalState.READY);
 
     const onSendEmail = async () => {
         try {
+            setModalState(EmailModalState.SENDING);
```
PanopticaRising, line 32: This should probably be a reducer, but when we make a generic state handler.

PanopticaRising (approved):

PR #440, opened 2021-02-18
Split query setting into a separate useEffect, added a guard around t…

PanopticaRising:
…he initial override of the problemId.


src/Assignments/SimpleProblemPage.tsx
```diff
@@ -146,7 +149,12 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
                 .keyBy('id')
                 .value();
             setProblems(problemDictionary);
-            setSelectedProblemId(_.sortBy(problems, ['problemNumber'])[0].id);
+
+            // If a selectedProblemId hasn't been set by the query parameter or is invalid, set it to the first id.
+            if (_.isNil(selectedProblemId) || !_.some(problems, ['id', selectedProblemId])) {
+                setSelectedProblemId(_.sortBy(problems, ['problemNumber'])[0].id);
```
red-tt-bear, line 155: is it guaranteed that problems has 1 element or should we use first?

red-tt-bear (approved):

PR #442, opened 2021-02-19
Feature/red 441 def export

red-tt-bear:


src/Courses/TopicSettings/TopicSettingsPage.tsx
```diff
@@ -191,7 +191,7 @@ export const TopicSettingsPage: React.FC<TopicSettingsPageProps> = ({topic: topi
     }, [topic]);
 
     const { getRootProps, getInputProps, open, isDragActive } = useDropzone({ onDrop,
-        accept: '.def',
+        accept: ['.def', '.rdef'],
```
PanopticaRising, line 194: Should this be `rtdf`?

PanopticaRising (changes requested):
Confirm extension change

src/Courses/TopicSettings/TopicSettingsPage.tsx
red-tt-bear: Nope

rtdf = rederly topic definition file - that is what I was calling our dump files
rdef = rederly def = rederly set definition file

I can change it if it's confusing, although it doesn't matter for this pr because I would change rtdf

PanopticaRising (approved):

PR #445, opened 2021-02-22
fix a couple of spelling issues

jeis4wpi:
would witespace clean up an spelling fixes be helpful?


red-tt-bear (approved):
Looks good, yeah those spelling errors are in logs so they got through

We have considered something like [prettier ](https://prettier.io/) for enforcing spacing but we haven't implemented/decided on what tool we want to use

PanopticaRising (approved):

PR #462, opened 2021-03-11
fix: Inverted the check for what types of topics to show the warning on.

PanopticaRising:


src/Courses/TopicSettings/HasEverBeenActiveWarning.tsx
```diff
@@ -7,7 +7,7 @@ interface HasEverBeenActiveWarningProps {
 }
 
 export const HasEverBeenActiveWarning: React.FC<HasEverBeenActiveWarningProps> = ({topic}) => {
-    if (topic.isExam() && topic.hasEverBeenActive()) return null;
+    if (!topic.isExam() && topic.hasEverBeenActive()) return null;
```
red-tt-bear, line 10: Shouldn't this be
```Typescript
    if (!topic.isExam() || !topic.hasEverBeenActive()) return null;
```
so homework or never active doesn't show the alert
the alert only shows if it is an exam and has been active

PanopticaRising: Ach, good catch.

PanopticaRising: This is why they say to not write negative conditionals.

red-tt-bear (approved):

PR #463, opened 2021-03-11
fix: Added a check to fill the current Version map with exactly one m…

PanopticaRising:
…issing version.

I'm not sure if the gradeInstanceId in this map matters, but keeping it consistent and will refactor later.

### DIFF
--- a/src/Courses/TopicGrades/GradeInfoHeader.tsx
+++ b/src/Courses/TopicGrades/GradeInfoHeader.tsx
@@ -120,10 +120,22 @@ export const GradeInfoHeader: React.FC<GradeInfoHeaderProps> = ({
                         } else {
                             // there should only be one studentGradeId in this case...
                             (map[wb.studentGradeId] || (map[wb.studentGradeId] = [])).push(parseInt(id, 10));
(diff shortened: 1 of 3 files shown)


src/Courses/TopicGrades/GradeInfoHeader.tsx
```diff
@@ -120,10 +120,22 @@ export const GradeInfoHeader: React.FC<GradeInfoHeaderProps> = ({
...
+                            if (wb.studentGradeId !== grade.id) {
+                                logger.error(`Workbooks for grade ${grade.id} are out of sync with the corresponding grade object.`);
+                            }
                         }
                         return map;
                     }, {} as { [k: number]: number[] });
+
+                    // In this case, the latest version has not been submitted yet.
+                    const currentVersionMapKeys = _.keys(currentVersionMap);
+                    if (grade.gradeInstances && currentVersionMapKeys.length === grade.gradeInstances.length - 1) {
```
red-tt-bear, line 132: either we should make this a `>` or add a check that if it is more than 1 off we console.warn
This way we don't get silent bugs that are hard to find

PanopticaRising:
Closing this as it was cherry-picked into #468

PR #472, opened 2021-03-12
Added logic to display "Grade Withheld" for any topics that do not co…

PanopticaRising:
…me back in the API call.

### DIFF
--- a/src/Assignments/SimpleProblemPage.tsx
+++ b/src/Assignments/SimpleProblemPage.tsx
@@ -448,11 +448,11 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
                 }
                 < div className="d-flex flex-row text-success">
                     <div className="d-flex flex-column flex-grow-1">Best Version Score</div>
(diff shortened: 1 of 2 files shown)


src/Assignments/SimpleProblemPage.tsx
```diff
@@ -437,19 +437,22 @@ export const SimpleProblemPage: React.FC<SimpleProblemPageProps> = () => {
                     Object.keys(problemScores).map(key => {
                         return (
                             <div className="d-flex flex-row" key={key}>
-                                <div className="d-flex flex-column flex-grow-1">{(key === 'total') ? 'Total for this attempt' : `Problem #${key}`}</div>
+                                <div className="d-flex flex-column flex-grow-1">
+                                    {(key === 'total') && 'Total points (this attempt)'}
```
red-tt-bear, line 441: I feel like I would prefer this logic outside of the tsx
Do the if statements above and set a message variable and then just use that

red-tt-bear (approved):

PR #474, opened 2021-03-19
Feature/add topic descriptions and feedback

PanopticaRising:


public/index.html
```diff
@@ -35,6 +35,20 @@
...
     <script src="/analytics.js"></script>
+
+
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.11.0/jquery.min.js"></script>
```
PanopticaRising, line 40: It would be cool to not load these on every page, or to load them all with `defer`, but I think that without lazy loading, we can't split them out meaningfully @red-tt-bear

PanopticaRising: We could add a hook to insert the script tag wherever Quill was used, not sure if that would get messy in any way.

src/Courses/TopicGrades/GradeFeedback.tsx
```diff
@@ -0,0 +1,74 @@
+import React, { useRef, useEffect } from 'react';
```
PanopticaRising, line 1: This is very similar to another Quill, but it's not controlled by a form, and isn't used in this PR.

red-tt-bear (approved):

PR #484, opened 2021-04-14
Feature/add topic feedback

PanopticaRising:


src/APIInterfaces/BackendAPI/Requests/CourseRequests.ts
```diff
@@ -671,6 +731,30 @@ export const postConfirmAttachmentUpload = async (options: PostConfirmAttachment
     }
 };
 
+export const postGenericConfirmAttachmentUpload = async (options: PostGenericConfirmAttachmentUploadOptions): Promise<AxiosResponse<BackendAPIResponse>> => {
```
red-tt-bear, line 734: Doesn't matter now but I feel like logic should be split out, this way only the api call itself is in here

red-tt-bear (approved):

red-tt-bear (approved):

PR #500, opened 2021-05-06
Gib/qa may 3rd bugfixes

PanopticaRising:


red-tt-bear:
Package lock has massive amounts of changes while package.json is unchanged

src/Components/Quill/QuillControlledEditor.tsx
```diff
@@ -233,6 +234,9 @@ export const QuillControlledEditor: React.FC<QuillControlledEditorProps> = ({onS
             {onSave && <Grid xs={12} style={{margin: '0% 1%'}}>
                 <Button disabled={disabled} fullWidth color='primary' variant='contained' onClick={onClickedSave}>Save Feedback</Button>
             </Grid>}
+            {!_.isNil(isDirty) && <Grid xs={12} style={{margin: '0% 1%'}}>
```
red-tt-bear, line 237: if is dirty isn't passed we hide the button?

red-tt-bear (changes requested):
package lock changes

src/Components/Quill/QuillControlledEditor.tsx
PanopticaRising: Yeah, there's `isDirty` which is only used if we're in a RHF and there's `onSave` which is the old behavior that will show a different button (line 235)

red-tt-bear (approved):

PR #504, opened 2021-05-11
feat: Added logic to detect if a file with an HEIC extension has a jp…

PanopticaRising:
…eg mime type, so default img handling can be used.

### DIFF
--- a/src/Components/Heic.tsx
+++ b/src/Components/Heic.tsx
@@ -34,6 +34,7 @@ export const Heic = forwardRef<HTMLImageElement, HeicProps>(
                 try {
                     const res = await fetch(url);
 
+                    // In Rederly architecture, HEIC files can be converted to JPEG files.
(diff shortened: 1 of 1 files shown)


src/Components/Heic.tsx
```diff
@@ -33,6 +33,12 @@ export const Heic = forwardRef<HTMLImageElement, HeicProps>(
             (async () => {
                 try {
                     const res = await fetch(url);
+
+                    if (res.headers.get('content-type') === 'image/jpeg') {
```
red-tt-bear, line 37: We should probably have a comment here as to why this is the way it is

Specifically, this is due to an assumption based on how the other components in the system behave

An heic is either an heic or a compressed jpeg

red-tt-bear (dismissed):

red-tt-bear (approved):

PR #510, opened 2021-05-17
fix: If the mime type is blank and the extension is heic, manually se…

PanopticaRising:
…t the Content Type on the upload.

### DIFF
--- a/src/APIInterfaces/AWS/Requests/StudentUpload.ts
+++ b/src/APIInterfaces/AWS/Requests/StudentUpload.ts
@@ -37,7 +37,7 @@ export const putUploadWork = ({
     try {
         return attachmentsAxios.put(presignedUrl.toString(), file, {
             headers: {
-                'Content-Type': file.type,
+                'Content-Type': (file.type === '' && file.name.toLowerCase().endsWith('.heic')) ? 'image/heic' : file.type,
             },
             onUploadProgress: onUploadProgress
         });


src/APIInterfaces/AWS/Requests/StudentUpload.ts
```diff
@@ -37,7 +37,7 @@ export const putUploadWork = ({
     try {
         return attachmentsAxios.put(presignedUrl.toString(), file, {
             headers: {
-                'Content-Type': file.type,
+                'Content-Type': (file.type === '' && file.name.toLowerCase().endsWith('.heic')) ? 'image/heic' : file.type,
```
red-tt-bear, line 40: @PanopticaRising is file.type always `''` for all browsers?

I think we can just keep this simple with `!file.type`

red-tt-bear: You're missing heif extension, looks like there are a bunch of extensions it could be
> .heif, .heifs; .heic, .heics; .avci, .avcs; .avif, .avifs
https://en.wikipedia.org/wiki/High_Efficiency_Image_File_Format

Probably worth it to just add the library

PanopticaRising: Do you have a library in mind?

red-tt-bear: No but I googled 🤣 
https://www.npmjs.com/package/mime-types

red-tt-bear:
```Typescript
file.type ? file.type : (mime.lookup(file.name) || file.type)
```

I checked and null, undefined, and empty string work as expected

This way we don't need to update this every time we have a weird case like this

PR #517, opened 2021-05-20
fix: Temporary patch for a state issue that causes the problemState t…

PanopticaRising:
…o revert, during exams, to their previous values.

This is a brittle fix because I'm not updating the Grade when I update the Grade Instance because the backend call does not return updated Grade information on save.
Right now, this shouldn't affect anything, but in the future, it could be confusing with a temporary lapse in synchronicity.


red-tt-bear:
Will review in stage pr

PR #524, opened 2021-05-29
RED-415 Added export functionality

red-tt-bear:


src/Courses/CourseDetailsTabs/CourseDetailsTab.tsx
```diff
@@ -113,14 +115,32 @@ export const CourseDetailsTab: React.FC<CourseDetailsTabProps> = ({ course, load
                 {userType !== UserRole.STUDENT &&
                     <div
                         style={{
-                            marginLeft: 'auto'
+                            marginLeft: 'auto',
+                            display: 'flex',
                         }}
                     >
+                        { inEditMode &&
```
PanopticaRising, line 122: Is there a reason to only show the Course Archive while in Edit Mode?

PanopticaRising (approved):

src/Courses/CourseDetailsTabs/CourseDetailsTab.tsx
red-tt-bear: Import is only in edit mode, I don't want it to just sit there all the time since it is not used often

PR #530, opened 2021-06-14
Feature/add superadmin check user

PanopticaRising:


src/APIInterfaces/BackendAPI/Requests/UserRequests.ts
```diff
@@ -207,6 +213,16 @@ export const impersonate = async ({
     }
 };
 
+export const superAdminGetUser = async ({
+    email
+}: {email: string}): Promise<BackendAPIResponse> => {
```
red-tt-bear, line 218: This generic should specify user

src/SuperAdmin/CheckUserInfo.tsx
```diff
@@ -0,0 +1,53 @@
...
+            user === null ? 
+                <>No user found</> : 
+                <table>
+                    {_.keys(user).map((key) => <tr key={key}>
+                        <td style={{border: '1px solid black'}}>
+                            {key}
+                        </td>
+                        <td style={{border: '1px solid black'}}>
+                            {key === 'paidUntil' || key === 'updatedAt' ? 
+                                `${moment(_.get(user, key))} (${moment(_.get(user, key)).calendar()})` :
```
red-tt-bear, line 43: I'd probably have made this a function with brackets so you could have a variable, it would look cleaner to not repeat `moment(_.get(user,key))`

src/SuperAdmin/CheckUserInfo.tsx
```diff
@@ -0,0 +1,53 @@
...
+        {user !== undefined && (
+            user === null ? 
+                <>No user found</> : 
+                <table>
+                    {_.keys(user).map((key) => <tr key={key}>
+                        <td style={{border: '1px solid black'}}>
+                            {key}
+                        </td>
+                        <td style={{border: '1px solid black'}}>
+                            {key === 'paidUntil' || key === 'updatedAt' ?
```
red-tt-bear, line 42: Does this work, you have to be careful with ternary and more complicated logic, the conditions should be wrapped in parenthesis

It might work but it seems dangerous

tommy-lettieri (approved):

